PR #29, opened 2018-07-25
Uni kde

aliciasun:
Added a KDEUnivariate class. It has the same functions as documented in the GaussianUnivariate class.


copulas/univariate/KDEUnivariate.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class KDEUnivariate(UnivariateDistrib):
+    """ Use Kernel Density Estimation to infer univariate distribution """
+
+    def __init__(self):
+        super(KDEUnivariate, self).__init__()
```
ManuelAlvarezC: Please, add proper documentation. This include docstrings on a module and class level, also, on each method.

For methods, docstrings should contain at least:
- One line explanation of what the method does.
- Expected input parameters, with it's type and default values if any.
- Expected output format and type.
- Larger explanation of the side effects, exceptions raises, assumptions and/or restrictions to be called. In methods that implement mathematical procedures (pdf, cdf, ...) It could be a good idea to have the matematical formulation of what is being implemented in code.
- An small example of use (only on non-trivial cases)
For example:
```python
def add_nums(x,y):
    """Return the sum of two values.

    Args:
        :param x: First value to be sumed.
        :param y: Second value to be sumed.
        :type x: int or float
        :type y: int or float

    Returns:
        sum: int or float with the sum of the values.
    
    This function will sum the two values passed as arguments and return the result.
    It will raise TypeError if the arguments can not be summed.
    If one of the values if a float the result will also be a float.

   Example: 
        >>> add_nums(1, 1)
        2
    """
    return x + y
```
You can have a look at [PEP257](https://www.python.org/dev/peps/pep-0257/) if you need further indication on doing so.

copulas/univariate/KDEUnivariate.py
```diff
@@ -0,0 +1,39 @@
...
+        self.column = None
+        self.param = None
+
+    def fit(self, column):
+        self.column = column
```
ManuelAlvarezC: What is `self.column` used for? I see no any other reference to it anywhere in the class.

copulas/univariate/KDEUnivariate.py
```diff
@@ -0,0 +1,39 @@
+import logging
```
ManuelAlvarezC: This comment doesn't belong to this concrete line, but to the whole module: There is not a single unittest.

There should be at least one single test for each method, making it run in controlled, standard conditions and checking the behavior is the expected. There should be also tests covering corner cases, know issues and limitations of the functions.

In the case of this module, I'll start with an schema like this:
```python
from unittest import TestCase

from copulas.univariate.KDEUNivariate import KDEUNivariate


class TestKDEUnivariate(TestCase):

    def test___init__(self):
        """On init, param and column attributes are set to None."""
        pass

    def test_fit(self):
        """On fit, kde model is instantiated with passed data"""
        pass

    def test_fit_empty_data(self):
        """If fitting kde model with empty data it will crash""" # Will it?
        pass

    def test_get_pdf(self):
        """get_pdf evaluates with the model"""
        pass

    def test_get_pdf_invalid_value(self):
        """Evaluating an invalid value will crash"""
    ...
```

ManuelAlvarezC:
Also, please, update your local master branch ( which now builds fine) and merge it here, and after that fix any build issues that may arise.

kveerama:
@aliciasun a few recommendations: perhaps change the name the PR to be descriptive. Also I liked your comparisons between using scipy vs. stats toolbox for this. Could we add some description about that here?

copulas/univariate/KDEUnivariate.py
aliciasun: It is not used in this class. But it might be used in other classes and it is easier to retrieve the data.

copulas/univariate/KDEUnivariate.py
```diff
@@ -0,0 +1,39 @@
...
+    """ Use Kernel Density Estimation to infer univariate distribution """
+
+    def __init__(self):
+        super(KDEUnivariate, self).__init__()
+        self.column = None
+        self.param = None
+
+    def fit(self, column):
+        self.column = column
+        model = gaussian_kde(column)
```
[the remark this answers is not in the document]
aliciasun: Thanks! Will change that.

copulas/univariate/KDEUnivariate.py
ManuelAlvarezC: This may not be the best of reasons:
- If another class is calling this one, they will have access to columns, as they'll have passed it on the call to `fit`.
- As a general rule classes shouldn't have attributes that they are no expecting to use. There are expections, but usually is this way

copulas/univariate/KDEUnivariate.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self):
+        super(KDEUnivariate, self).__init__()
+        self.model = None
+
+    def fit(self, column):
+        """Fit Kernel density estimation to an list of values.
+
+        Args:
+            :param column: list of datapoints to be estimated from.
+            :type column: 1-D array
```
ManuelAlvarezC: type should be specified as the class, ie `np.ndarray` or `pd.Series`

tests/univariate/test_KDEUnivariate.py
```diff
@@ -0,0 +1,61 @@
...
+    def test___init__(self):
+        """On init, model are set to None."""
+        self.kde = KDEUnivariate()
+        assert self.kde.model is None
+
+    def test_fit_uniform(self):
+        """On fit, kde model is instantiated with passed data"""
+        self.kde = KDEUnivariate()
+        self.kde.fit([1, 2, 3, 4, 5])
+        assert self.kde.model is not None
```
ManuelAlvarezC: Here you are not testing that the model is instantiated with data, you are just looking that the attribute `model` of the instance is not `None`.

You should check that model is indeed a  `scipy.stats.gaussian_kde` instance. And check also the attribute `dataset`of the model is the same array you passed to `fit`

tests/univariate/test_KDEUnivariate.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_get_pdf(self):
+        """get_pdf evaluates with the model"""
+        self.setup_norm()
+        x = self.kde.get_pdf(0.5)
+        expected = 0.35206532676429952
+        self.assertAlmostEquals(x, expected, places=1)
+
+    def test_get_pdf_invalid_value(self):
+        """Evaluating an invalid value will crash"""
```
ManuelAlvarezC: This test is missing

tests/univariate/test_KDEUnivariate.py
```diff
@@ -0,0 +1,61 @@
...
+"""Tests for `univariate` package."""
+
+from unittest import TestCase
+
+import numpy as np
+
+from copulas.univariate.KDEUnivariate import KDEUnivariate
+
+
+class TestKDEUnivariate(TestCase):
```
ManuelAlvarezC: This is just a minor styling issue but helps understanding the code: When possible, split each test in three blocks like:
```python
class TestKDEUnivariate(TestCase):

    def test_example(self):
        """On special circumstances, model may sample values."""
        # Setup
        model = KDEUnivariate()
        model.fit([1,2,3,4,5])

        # Run
        result = model.sample()

        # Check 
        assert result
```

tests/bivariate/test_copula_selection.py
```diff
@@ -20,6 +20,7 @@ def test_snippet(self):
         c0 = Copula('clayton')
         c0.fit(U, V)
         result = c0.get_cdf()([0, 0.1, 0.2], [0, 0.1, 0.8])
+        print(result)
```
ManuelAlvarezC: Please, delete this line, statements used for developing should never be committed ( `print`s, `ipdb`s, etc, etc). If you intend to output a message to the user, which is not the case here, use a logger.

copulas/bivariate/copulas.py
```diff
@@ -312,9 +312,12 @@ def compute_empirical(u, v):
     def select_copula(U, V):
```
ManuelAlvarezC: Changes on this file doesn't belong to this PR

PR #41, opened 2018-08-09
4 add unittest univariate

### LINKED ISSUE #4: add tests for univariate models

ManuelAlvarezC:
Added tests for univariate gaussian copula.
Resolve #4


tests/univariate/test_gaussian.py
```diff
@@ -0,0 +1,169 @@
...
+        column = [-1, 0, 1]
+        copula.fit(column)
+
+        # Run
+        result = copula.sample(1000000)
+
+        # Check
+        assert len(result) == 1000000
+        assert (np.mean(result) - copula.mean) < 10E-3
+        assert (np.std(result) - copula.std) < 10E-3
```
aliciasun: I think it is better to use `self.assertAlmostEquals` or absolute value here since `np.mean(result)` could be much smaller than `copula.mean` but the assert above is still true.

ManuelAlvarezC: Right! I'll add the absoulte value.

Issue #11, opened 2018-05-18
Improve data management


ManuelAlvarezC:
Don’t store iris dataset on repo,  download it using sklearn.load_data

echo66:
Question: if you are going to include sklearn as a dependency, wouldn't it be a good idea to replace the current KDE implementation with the one from sklearn?

ManuelAlvarezC:
Hi @echo66, and thanks for your suggestion.

Would you mind elaborating a little why do you think the KDE implementation from `sklearn` is better than the one from `scipy`?

csala:
This is not related to the current project status any more.

PR #42, opened 2018-08-10
36 separate bivariate copulas

### LINKED ISSUE #36: Separate Bivariate Copulas
Bivariate Copulas should be separated into a class for each. Also the copula selector class `copulas.bivariate.bv_copula` should use inheritance to select one or another, instead of if statements

ManuelAlvarezC:
Bivariate Copulas have been separated into a model of its own. 
A mutable base class has been created to switch between them.
Vines and Trees have been adapted to this new API.
Resolve #36


copulas/utils.py
```diff
@@ -19,6 +19,7 @@
 RAW_EXT = '.raw.csv'
 SYNTH_EXT = '.synth.csv'
 TRANS_EXT = '.trans.csv'
+EPSILON = np.finfo(np.float32).eps
```
aliciasun: I think most part of this file is not being used anymore. The classes implemented in this file is used for the SDV project. Would you mind removing them?

copulas/bivariate/base.py
```diff
@@ -0,0 +1,159 @@
...
+                copula_type = CopulaTypes[copula_type.upper()]
+            else:
+                raise ValueError('Invalid copula type {}'.format(copula_type))
+
+        for subclass in cls.subclasses():
+            if subclass.copula_type is copula_type:
+                return super(Bivariate, cls).__new__(subclass)
+
+    def __init__(self, copula_type=None):
+        """ initialize copula object """
```
aliciasun: Please add documentation and example for the init method.  I think it can either be instantiated with a lower case string or a member in CopulaType class?

copulas/bivariate/base.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def get_cdf(self):
+        """ returns cdf of model """
+        raise NotImplementedError
+
+    def copula_sample(self, v, c, amount):
+        raise NotImplementedError
+
+    def sample(self, amount):
+        """ returns a new data point generated from model
```
aliciasun: Amount is a little bit confusing. Maybe use num_sample as the variable name?

copulas/bivariate/frank.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+
+        def debye(t):
+            return t / (np.exp(t) - 1)
+
+        debye_value = integrate.quad(debye, sys.float_info.epsilon, alpha)[0] / alpha
+        return 4 * (debye_value - 1) / alpha + 1 - tau
+
+    def copula_sample(self, v, c, amount):
+        u = np.empty([1, amount])
```
aliciasun: amount is just the length of v. Just do `amount = len(v)` instead of passing amount as an argument.

copulas/bivariate/clayton.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def tau_to_theta(self):
+        if self.tau == 1:
+            theta = 10000
+        else:
+            theta = 2 * self.tau / (1 - self.tau)
+        return theta
+
+    def copula_sample(self, v, c, amount):
+        return self.get_ppf()(c, v, self.theta)
```
aliciasun: Same here. amount can be removed as an argument.

copulas/bivariate/base.py
ManuelAlvarezC: Done.

Not only that, right now we have:
```python
>>> Bivariate(CopulaTypes.FRANK).__class__  # Using CopulaType
copulas.bivariate.frank.Frank

>>> Bivariate('frank').__class__ # Using strings
copulas.bivariate.frank.Frank
```
But thanks to CopulaType being an `enum.Enum`, we can also acces its members by both name and value:

```python
>>> Bivariate(CopulaTypes['FRANK']).__class__
copulas.bivariate.frank.Frank

>>> Bivariate(CopulaTypes(1)).__class__
copulas.bivariate.frank.Frank
```

This is useful in scenarios when we want to instantiate a Copula whose type won't be known until runtime, like when doing the copula selection.

copulas/bivariate/base.py
ManuelAlvarezC: I don't think in this context `amount` is confusing.

If we where in an scenario where our function have multiple arguments that refers to amounts, it shoudl, of course, be added the concept in the argument name.

But in this case we have a single argument function, and I don't think `amount` could be thought as something different the amount of samples to produce.

copulas/bivariate/clayton.py
ManuelAlvarezC: Sure.

copulas/bivariate/frank.py
ManuelAlvarezC: Yep, that was a miss. Thanks for catching it!

copulas/utils.py
ManuelAlvarezC: I had planned doing that along issue #11. After move the test data functions out of here, I'll delete this file.

PR #43, opened 2018-08-12
Add unittest  for likelihood

aliciasun:


copulas/multivariate/tree.py
```diff
@@ -299,13 +276,12 @@ def _build_first_tree(self):
...
-            name, theta = Bivariate.select_copula(
-                self.u_matrix[:, edge[0]], self.u_matrix[:, edge[1]])
-
-            tau_matrix = self.tau_matrix[edge[0], edge[1]]
-            new_edge = Edge(itr, edge[0], edge[1], tau_matrix, name, theta)
-            new_edge.parent.append(edge[0])
-            new_edge.parent.append(edge[1])
+            name, theta = Bivariate.select_copula(self.u_matrix[:, edge[0]],
+                                                  self.u_matrix[:, edge[1]])
+            left = min(edge[0], edge[1])
```
ManuelAlvarezC: If you intend to set the min value in 'edge` as left, and the max as right, you can simply do.

```python
left, right = sorted(edge)
```
You can even do it on Edge.__init__, accepting just two nodes in an iterable, and do the assignment on the class instantiation, and avoid the sorting before creating the Edge instance. Something like:

```python
class Edge(object):
    def __init__(self, nodes, *args, **kwargs):

        self.left, self.right = sorted(nodes)

edge = Edge([7,4])
assert edge.L == 4
assert edge.R ==7 
```

copulas/multivariate/tree.py
```diff
@@ -414,19 +373,57 @@ def is_adjacent(self, another_edge):
         return (self.L == another_edge.L or self.L == another_edge.R or
                 self.R == another_edge.L or self.R == another_edge.R)
 
-    def get_likelihood(self, U):
+    @staticmethod
+    def sort_edge(edge1, edge2):
```
ManuelAlvarezC: Considering that you would always use two instances of `Edge` to call `sort_edge`, and also that the input order should no matter, as we are trying to order them, wouldn't it make more sense to make this `staticmethod`  a regular instance method, consider `edge1` to be `self` and call it from one of the instances to be sorted?

```python
first, second = edge1.sort(edge2)
first, second = Edge.sort_edge(edge1, edge2)
```

tests/copulas/multivariate/test_direct_tree.py
```diff
@@ -0,0 +1,52 @@
+import logging
+from unittest import TestCase
+
+import numpy as np
+import pandas as pd
+
+from copulas.multivariate.tree import DirectTree
+from copulas.univariate.kde import KDEUnivariate
+
+LOGGER = logging.getLogger(__name__)
```
ManuelAlvarezC: Same as above.

tests/copulas/multivariate/test_regular_tree.py
```diff
@@ -0,0 +1,49 @@
+import logging
+from unittest import TestCase
+
+import numpy as np
+import pandas as pd
+
+from copulas.multivariate.tree import RegularTree
+from copulas.univariate.kde import KDEUnivariate
+
+LOGGER = logging.getLogger(__name__)
```
ManuelAlvarezC: Same as above.

tests/copulas/multivariate/test_vine.py
```diff
@@ -1,28 +1,42 @@
...
-        sample_r = self.rvine.sample()
-        sample_d = self.dvine.sample()
-        sample_c = self.cvine.sample()
+        dvalue = self.dvine.get_likelihood(uni_matrix)
+        self.assertAlmostEquals(dvalue, -5.18387, places=3)
 
-        self.assertEquals(len(sample_r), 4)
-        self.assertEquals(len(sample_d), 4)
-        self.assertEquals(len(sample_c), 4)
+    # def test_sample(self):
```
ManuelAlvarezC: Is there a reason to have this test commented out?

tests/copulas/multivariate/test_center_tree.py
```diff
@@ -0,0 +1,53 @@
...
+            self.u_matrix[:, count] = [uni.get_cdf(x) for x in self.data[col]]
+            count += 1
+        self.trees = []
+        self.trees.append(CenterTree(0, 4, self.tau_mat, self.u_matrix))
+
+    def test_first_tree(self):
+        self.assertEquals(self.trees[0].edges[0].L, 0)
+
+    def test_first_tree_likelihood(self):
+        print(self.trees[0])
```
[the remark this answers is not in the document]
aliciasun: I ran it with pytest -s to get the stdout. What is the correct way if I want to debug when running using pytest (sth like print statement or debugger in a regular program)?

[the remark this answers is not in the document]
csala: @aliciasun The problem here is not about the debugging itself which you can do in any way you like (I personally put `ipdb.set_trace()` statements to go into interactive debuggin, but using print statemtents can also work).
The problem is, instead, about the fact that you left there the things that you are using for debugging (i.e. the print statement).

Think about it this way: Unit tests are somehting automated that needs to be run and validated by a machine (in this case Travis), so you can rely on looking at an output with your eyes to tell whether the test is good or not.

In other words: If this print has been put there to validate something by looking at the outputs, it should  be converted to the corresponding assertions for travis to be able to detect the errors.
Otherwise, if the print statement has been put there because you used it while debugging a strange behaviour that you could not understand otherwise, it should be removed.

copulas/multivariate/tree.py
```diff
@@ -63,6 +64,24 @@ def _get_constraints(self):
...
 
+    def _sort_tau_by_y(self, y):
+        """Sort tau matrix by dependece with variable y
+
+        Args:
+            :param y: index of variable of intrest
+            :type y: int
+        """
+        # first column is the variable of interest
+        np.fill_diagonal(self.tau_matrix, np.NaN)
```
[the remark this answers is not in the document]
aliciasun: Oh yeah that's a good point. I changed it so that it only changes one column.

copulas/multivariate/tree.py
aliciasun: I changed the arguments such that it could be passed in a list of edges. Thought this would be more convenient.

PR #52, opened 2018-08-23
Issue 48: Fix means in gaussian copula sample method

### LINKED ISSUE #48: Fix means in gaussian copula sample method
The means used in the sample method of a copulas.multivariate.GaussianMultivariate instance should be a list of 0s with the length of the number of columns in the distribution. This is on lines 85 and 89 of the https://github.com/DAI-Lab/Copulas/blob/master/copulas/multivariate/gaussian.py.

amontanez24:
Resolves #48. This PR changes the sample method of the  copulas.multivariate.GaussianMultivariate to use an array of 0s as the means instead of the means of the columns. This is the correct way to form a Gaussian Copula according to https://en.wikipedia.org/wiki/Copula_(probability_theory)#Gaussian_copula


copulas/multivariate/gaussian.py
```diff
@@ -93,15 +93,12 @@ def func(*args):
 
     def sample(self, num_rows=1):
         res = {}
-
-        # clean up means
-        clean_mean = np.nan_to_num(self.means)
+        means = [0] * len(self.means)
```
ManuelAlvarezC: This could be replaced with `np.zeros(len(self.means)))`, that it's more clear than multiplying a list.

tests/copulas/multivariate/test_gaussian_copula.py
```diff
@@ -22,3 +23,19 @@ def test_deprecation_warnings(self):
...
+    def test_sample(self):
+        copula = GaussianMultivariate()
+        data = pd.DataFrame([[-1, 0, 1], [1, -1, 0], [0, 1, -1]])
+        copula.fit(data)
+
+        # Run
+        result = copula.sample(1000000)
+
+        # Check
+        assert len(result) == 1000000
```
ManuelAlvarezC: Change this assertion to:
```python
assert result.shape == (1000000, 3)
```
This is because, the `for` loop below won't raise any error if `error.shape[1] < data.shape[1]`

tests/copulas/multivariate/test_gaussian_copula.py
```diff
@@ -22,3 +23,19 @@ def test_deprecation_warnings(self):
...
+        for i in range(result.shape[1]):
+            data_mean = np.mean(data.loc[:, i])
+            result_mean = np.mean(result.loc[:, i])
+            assert abs(data_mean - result_mean) < 10E-3
+            assert abs(data_mean - result_mean) < 10E-3
```
ManuelAlvarezC: This line is repeated, and the checks for the standard deviation are missing.

tests/copulas/multivariate/test_gaussian_copula.py
```diff
@@ -22,3 +23,19 @@ def test_deprecation_warnings(self):
             # Check
             assert len(warns) == 0
             assert len(result) == 10
+
+    def test_sample(self):
```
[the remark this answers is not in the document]
ManuelAlvarezC: I can run this test on master and it passes. That means that this test is not able to capture the change of behavior is trying to test.

PR #54, opened 2018-08-28
Issue 33: Add methods to load and save copulas

### LINKED ISSUE #33: Add methods to load/save the intenal state of a copula.
We need to add a `save`and `load` methods to replicate the internal state of an copula

ManuelAlvarezC:
This PR adds the functionality to load and save the internal state of a copula in a file.
To do so, the following methods have been added to the classes:
- `to_dict`: Returns a `dict` containing the required attributes to replicate an object.
- `from_dict`: Sets the internal state according to the passed `dict`.
- `save`: Stores the contents of `to_dict` in the specified file as a JSON.
- `load`: Returns an instance with the attributes loaded from a file or `dict`.

In classes where there were similar methods, those have been rename to keep consistency in the API.
Resolves #33


copulas/bivariate/base.py
```diff
@@ -76,12 +77,19 @@ def fit(self, U, V):
...
+        }
 
-    def set_params(self, **kwargs):
-        for key, value in kwargs.items():
-            setattr(self, key, value)
+    @classmethod
+    def from_dict(cls, **kwargs):
+        instance = cls(kwargs['copula_type'])
+        instance.theta = kwargs['theta']
+        instance.tau = kwargs.get('tau')
```
csala: This should be taken without get: it must be given

copulas/bivariate/base.py
```diff
@@ -76,12 +77,19 @@ def fit(self, U, V):
...
+            'copula_type': self.copula_type.name,
+            'theta': self.theta,
+            'tau': self.tau
+        }
 
-    def set_params(self, **kwargs):
-        for key, value in kwargs.items():
-            setattr(self, key, value)
+    @classmethod
+    def from_dict(cls, **kwargs):
```
csala: This should expect a variable called `copula_dict` or similar, which should be a dict, instead of kwargs.

copulas/bivariate/base.py
```diff
@@ -199,3 +207,17 @@ def compute_empirical(u, v):
...
+
+    def save(self, filename):
+        """Save the internal state of a copula in the specified filename."""
+        content = self.to_dict()
+        with open(filename, 'w') as f:
+            json.dump(content, f)
+
+    @classmethod
+    def load(cls, copula_path):
+        """Creates a new instance from a file or dict."""
```
csala: Write the docstrings in imperative mode instead of third person singluar

copulas/multivariate/gaussian.py
```diff
@@ -37,8 +37,25 @@ def __str__(self):
...
+        if distrib_map is not None:
+            for key in distrib_map:
+                # this isn't fully working yet
+                self.distribs[key] = distrib_map[key](data[key])
+
+        else:
+            for key in keys:
+                self.distribs[key] = GaussianUnivariate()
+                self.distribs[key].fit(data[key])
+        self.cov_matrix, self.means, self.distribution = self._get_parameters(data)
```
csala: Please add an empty line above this one

copulas/multivariate/tree.py
```diff
@@ -176,6 +177,34 @@ def __str__(self):
...
+    def to_dict(self):
+        return {
+            'level': self.level,
+            'previous_tree': self.previous_tree.tolist(),
+            'n_nodes': self.n_nodes,
+            'edges': [edge.to_dict() for edge in self.edges],
+            'tau_matrix': self.tau_matrix.tolist()
+        }
+
+    def from_dict(self, **kwargs):
```
csala: replace `**kwargs` with `copula_dict`

copulas/multivariate/tree.py
```diff
@@ -176,6 +177,34 @@ def __str__(self):
...
+        """Set attributes with provided values."""
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls, tree):
```
csala: This needs to be adapted like the bivariate copula.

copulas/multivariate/tree.py
```diff
@@ -418,9 +446,55 @@ def get_likelihood(self, uni_matrix):
             right_ing = list(self.D - self.parents[1].D)[0]
             left_u = uni_matrix[self.L, left_ing]
             right_u = uni_matrix[self.R, right_ing]
-        cop = Bivariate(self.name)
-        cop.set_params(theta=self.theta)
+        cop = Bivariate.from_dict(copula_type=self.name, theta=self.theta)
```
csala: Please add an empty line above this one

csala: Also, here, instead of calling from_dict, create the instance of the copula first and then set the theta value directly as an attribute:

```
cop = Bivariate(self.name)
cop.theta = self.theta
```

PR #62, opened 2018-10-10
Added seed for random numbers generators

echo66:
Right now, when using the sampling methods, in order to get the same result everytime, we need to do the following things:

1. invoke `np.random.seed(seed_value)`
2. invoke `random.set_state(random_state_tuple)`

outside of the function being called (i.e. `sample`). This is what people call a "leaky abstraction", in software engineering. An easy way to cope with this is to include the seed as a parameter in the `sample` method. Usually, we just take that seed, use it as an input for something like `sklearn.check_random_state` and we receive a random numbers generator as an output, using it instead of `numpy.random`. But since the code for some classes is using functions that allow us to pass the seed or random numbers generator instantiated by us, we need to be a little bit more sophisticated by using `numpy.random.get_state`, `numpy.random.set_state`, `random.getstate` and `random.setstate`.

### DIFF
--- a/copulas/bivariate/base.py
+++ b/copulas/bivariate/base.py
@@ -194,14 +194,14 @@ class Bivariate(object):
         if self.tau > 1 or self.tau < -1:
             raise ValueError("The range for correlation measure is [-1,1].")
             
-        s, keys, pos, has_gauss, cached_gaussian = np.random.get_state()
+        s = np.random.get_state()
         
         np.random.seed(random_state)
 
         v = np.random.uniform(0, 1, n_samples)
         c = np.random.uniform(0, 1, n_samples)
         
-        np.random.set_state((s, keys, pos, has_gauss, cached_gaussian))
+        np.random.set_state(s)
 
         u = self.percent_point(c, v)
         return np.column_stack((u, v))
--- a/copulas/bivariate/base.py
+++ b/copulas/bivariate/base.py
@@ -188,6 +188,8 @@ class Bivariate(object):
         Args:
             n_samples: `int`, amount of samples to create.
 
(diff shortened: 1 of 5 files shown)


echo66:
I can't understand what is failing in Travis. All those errors seem related to docstrings and rst files I didn't touch.

csala:
Thanks for you contribution @echo66 !

Indeed, being able to pin the random seed within the sampling methods itself seems to be an interesting addition, so we really appreciate your interest in contributing these changes.

Unfortunately, I'm afraid that your PR does not adhere to our contributing guidelines, which you can find [here](https://dai-lab.github.io/Copulas/contributing.html)

Briefly, I suggest the following approach:

* Cancel the current PR (temporary).
* Open a GitHub issue explaining what you would like to see changed, why, and how you would implement it. We would love to share our impressions there and be able to briefly discuss any possible discrepancy about the ideal way to implement the changes.
* Once we have agreed on the implementation, please make sure that:
    * Your implementation corresponds to what has been discussed in the issue.
    * There are unit tests that explicitly validate the changes associated with the issue
    * The changes are well documented, especially if there are API changes, like the new method arguments that this PR seems to add.
    * All the tests and code styling checks pass successfully in all platforms (running `make test-all`).
* Once all this is validated, please issue a new PR and we'll gladly review and merge it as soon as we can.

csala:
Regarding the Travis build failure, it seems to be because some of the changed files do not pass the `flake8` validation.

You check this locally by running `make lint`, which will show a nice report about whether the code styling checks pass or not and, if they don't, what things need to be changed.

echo66:
Thanks for the feedback! I will do as you told.

PR #76, opened 2019-01-07
Issue 55: Add distribution argument to Gaussian copula

### LINKED ISSUE #55: Gaussian Copula should take argument for distribution
https://github.com/DAI-Lab/Copulas/blob/e1f966a1775e63bc7fadf1b13bb0d2567a30ce78/copulas/multivariate/gaussian.py#L74

If no distribution map is given, the Gaussian Copula uses the GaussianUnivariate class as the default distribution. It should instead take a distribution in as an argument and use that as the default.

ManuelAlvarezC:
The following have been done:
- Add distribution to `__init__` arguments and store it in the distribution attribute. 
- On fit time, distribution is imported and used.
- Update old tests, added new tests to check everything works as expected.

Resolves #55

### DIFF
--- a/copulas/multivariate/gaussian.py
+++ b/copulas/multivariate/gaussian.py
@@ -4,9 +4,9 @@ import numpy as np
 import pandas as pd
 from scipy import integrate, stats
 
-from copulas import get_qualified_name
+from copulas import get_qualified_name, import_object
 from copulas.multivariate.base import Multivariate
-from copulas.univariate import GaussianUnivariate, Univariate
+from copulas.univariate import Univariate
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


copulas/multivariate/gaussian.py
```diff
@@ -4,22 +4,27 @@
...
+        distribution (str): Full qualified name of the class to be used as distribution.
+    """
+
+    def __init__(self, distribution=None):
         super().__init__()
 
         self.distribs = {}
         self.covariance = None
         self.means = None
+        self.distribution = distribution or 'copulas.univariate.gaussian.GaussianUnivariate'
```
csala: This would be better set as a module constant and as the default value for the argument:
```
DEFAULT_DISTRIBUTION = '...'
...
class ...:
    def __init__(self, distribution=DEFAULT_DISTRIBUTION):
        ...
```

PR #79, opened 2019-01-08
Revert "Issue 75: Fix error on vine sampling"

csala:
Reverts DAI-Lab/Copulas#78

### DIFF
--- a/copulas/multivariate/vine.py
+++ b/copulas/multivariate/vine.py
@@ -171,9 +171,7 @@ class VineCopula(Multivariate):
                                 condition = set(edge.D)
                                 condition.add(edge.L)
                                 condition.add(edge.R)
-
-                                visit_set = set(visited)
-                                visit_set.add(current)
+                                visit_set = set(visited).add(current)
 
                                 if condition.issubset(visit_set):
                                     current_ind = edge.index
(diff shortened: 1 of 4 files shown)


csala:
@ManuelAlvarezC I realized about a potential bug right after merging the previous PR, so I open this one just to discuss it.

If it is not a real bug, it can be discarded closed and discarded.

copulas/multivariate/vine.py
```diff
@@ -182,19 +180,19 @@ def sample(self, num_rows=1):
                     if current_ind != -1:
                         # the node is not indepedent contional on visited node
                         copula_type = current_tree[current_ind].name
-                        copula = Bivariate(CopulaTypes(copula_type))
-                        copula.theta = current_tree[current_ind].theta
-                        derivative = copula._partial_derivative
```
csala: @ManuelAlvarezC Shouldn't this be `partial_derivative` instead of `_partial_derivative`?

ManuelAlvarezC: @csala No, it should be`_partial_derivative` because it's scalar.It's true though that `Clayton` copula doesn't have that method, so we need to fix this.

csala: Ok, no problem @ManuelAlvarezC Let's close this to keep it cleaner and open an separated issue to fix this.

PR #88, opened 2019-02-14
Issue 82: Add ability to handle constant data with Univariate

### LINKED ISSUE #57: Review the single value column workaround
If the fit data has only one value, hence the std is 0, the current code has a workaround to avoid crashing later on when trying to crash:

https://github.com/DAI-Lab/Copulas/blob/f27d86d50db8e8e94c43bf6c0f17f650bffdd6ce/copulas/univariate/gaussian.py#L45

Wouldn't it be better to capture and implement this exceptional situation by just making all samples constant, equal to the mean?

### LINKED ISSUE #82: Add ability to handle constant data to univariate classes
* Copulas version: 0.2.1
* Python version: 3.6.1
* Operating System: Ubuntu 18.04.1 LTS

### Description

I was trying to `fit` a `copulas.univariate.kde.KDEUnivariate` with an array of constant data. I expected for it to work and be able to sample data ( altough I was supposing that the sampled values will be constant too).

### What I Did

```python
import numpy as np

from copulas.univariate import KDEUnivariate

X = np.array([1, 1, 1, 1])
kde = KDEUnivariate()
kde.fit(X)
```

and got the following traceback:
```LinAlgError                               Traceback (most recent call last)
<ipython-input-2-6d5d418eb1ce> in <module>
      5 X = np.array([1, 1, 1, 1])
      6 kde = KDEUnivariate()
----> 7 kde.fit(X)

~/Pythia/MIT/Copulas/copulas/univariate/kde.py in fit(self, X)
     27             raise ValueError("data cannot be empty")
     28 
---> 29         self.model = scipy.stats.gaussian_kde(X)
     30         self.fitted = True
     31 

~/.virtualenvs/copulas_mit/lib/python3.6/site-packages/scipy/stats/kde.py in __init__(self, dataset, bw_method, weights)
    206             self._neff = 1/sum(self._weights**2)
    207 
--> 208         self.set_bandwidth(bw_method=bw_method)
    209 
    210     def evaluate(self, points):

~/.virtualenvs/copulas_mit/lib/python3.6/site-packages/scipy/stats/kde.py in set_bandwidth(self, bw_method)
    538             raise ValueError(msg)
    539 
--> 540         self._compute_covariance()
    541 
    542     def _compute_covariance(self):

~/.virtualenvs/copulas_mit/lib/python3.6/site-packages/scipy/stats/kde.py in _compute_covariance(self)
    550                                                bias=False,
    551                                                aweights=self.weights))
--> 552             self._data_inv_cov = linalg.inv(self._data_covariance)
    553 
    554         self.covariance = self._data_covariance * self.factor**2

~/.virtualenvs/copulas_mit/lib/python3.6/site-packages/scipy/linalg/basic.py in inv(a, overwrite_a, check_finite)
    972         inv_a, info = getri(lu, piv, lwork=lwork, overwrite_lu=1)
    973     if info > 0:
--> 974         raise LinAlgError("singular matrix")
    975     if info < 0:
    976         raise ValueError('illegal value in %d-th argument of internal '

LinAlgError: singular matrix
```

There is an issue open (#57) to fix a workaround on `copulas.univariate.gaussian.GaussianUnivariate` that avoid this exact situation, could we generalize its solution on `copulas.univariate.base.Univariate` to be able to model and sample constant data with all univariate distributions?

ManuelAlvarezC:
Add the statistical functions (`cdf`, `pdf`, `ppf`, `sample`) of the degenerate(constant) distribution on  `copulas.univariate.base.Univariate` so can be used when fit with single-valued data.

Resolves #82 also Resolves #57

### DIFF
--- a/copulas/univariate/base.py
+++ b/copulas/univariate/base.py
@@ -105,16 +105,12 @@ class Univariate(object):
         if not self.fitted:
             raise NotFittedError("This model is not fitted.")
 
-    def check_constant_value(self):
-        if self.constant_value:
-            raise ValueError('This method is not available on constant distributions.')
-
     @staticmethod
     def _get_constant_value(X):
         """Checks if a Series or array contains only one unique value.
 
         Args:
-            X(pandas.Series or numpy.array): Array to check for constantness
+            X(pandas.Series or numpy.ndarray): Array to check for constantness
 
         Returns:
             (float or None): Return the constant value if there is one, else return None.
@@ -130,7 +126,7 @@ class Univariate(object):
             num_samples(int): Number of rows to sample
 
         Returns:
-            numpy.array: Sampled values. Array of shape (num_samples,).
+            numpy.ndarray: Sampled values. Array of shape (num_samples,).
         """
         return np.array([self.constant_value] * num_samples)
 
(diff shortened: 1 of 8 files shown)


copulas/univariate/base.py
```diff
@@ -100,3 +104,78 @@ def check_fit(self):
...
+    def _constant_sample(self, num_samples):
+        """Sample values for a constant distribution.
+
+        Args:
+            num_samples(int): Number of rows to sample
+
+        Returns:
+            numpy.ndarray: Sampled values. Array of shape (num_samples,).
+        """
+        return np.array([self.constant_value] * num_samples)
```
csala: I believe you could use `numpy.full` here.

copulas/univariate/gaussian.py
```diff
@@ -73,6 +78,9 @@ def cumulative_distribution(self, X):
             np.ndarray: Cumulative density for X.
         """
         self.check_fit()
+        if self.constant_value:
```
csala: Instead of checking for `self.constant_value` every time, I would just replace the instance method inside the `fit` method.

```
def fit():
    self.constant_value = self._get_constant_value(X)
    if self.constant_value is not None:
        self.cumulative_distribution = self._constant_cumulative_distribution
        # etc.
```

copulas/univariate/gaussian.py
```diff
@@ -44,11 +44,13 @@ def fit(self, X):
...
         if isinstance(X, (pd.Series, pd.DataFrame)):
             self.name = X.name
-        else:
-            self.name = None
 
-        self.mean = np.mean(X)
-        self.std = np.std(X) or 0.001
+        self.constant_value = self._get_constant_value(X)
+
+        if not self.constant_value:
```
csala: Remember that `bool(0) == False`.
This should be an explicit check for `None`:  `if self.constant_value is not None`

PR #100, opened 2019-08-12
Issue 99: Remove constant_value key in distribution serialization

### LINKED ISSUE #99: Improve serialization of univariate distributions
The current implementation of the `to_dict` and ` form_dict` methods exposes the internal attribute `constant_value` and returns different values depending on its value. 

We should change the implementation to make that all fitted distributions of the same class have the exact same serialization independently of the value of the `constant_value` attribute, and also have the attribute `constant_value` implicitly codified in the serialization.

ManuelAlvarezC:
Resolves #99 

Removed the key `constant_value` on the serialization of univariate distributions.

### DIFF
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -82,21 +82,37 @@ class GaussianKDE(ScipyWrapper):
         instance = cls()
 
         instance.fitted = copula_dict['fitted']
-        instance.constant_value = copula_dict['constant_value']
 
-        if instance.fitted and not instance.constant_value:
-            instance.model = scipy.stats.gaussian_kde([-1, 0, 0])
+        if instance.fitted:
 
-            for key in ['dataset', 'covariance', 'inv_cov']:
-                copula_dict[key] = np.array(copula_dict[key])
+            covariance = copula_dict['covariance']
+            if not isinstance(covariance, list):
+                instance.constant_value = covariance
 
-            attributes = ['d', 'n', 'dataset', 'covariance', 'factor', 'inv_cov']
-            for name in attributes:
-                setattr(instance.model, name, copula_dict[name])
+            if instance.constant_value is None:
+                instance.model = scipy.stats.gaussian_kde([-1, 0, 0])
+
+                copula_dict['covariance'] = np.array(covariance)
+                for key in ['dataset', 'inv_cov']:
+                    copula_dict[key] = np.array(copula_dict[key])
+
+                attributes = ['d', 'n', 'dataset', 'covariance', 'factor', 'inv_cov']
+                for name in attributes:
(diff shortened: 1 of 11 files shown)


copulas/univariate/truncnorm.py
```diff
@@ -19,12 +19,18 @@ class TruncNorm(ScipyWrapper):
...
-        self.std = X.std()
+        self.constant_value = self._get_constant_value(X)
+        if self.constant_value is None:
+            min_ = X.min() - EPSILON
+            max_ = X.max() + EPSILON
+            mean = X.mean()
+            std = X.std()
 
-        super().fit(X, min_, max_)
+            super().fit(X, min_, max_, loc=mean, scale=std)
```
csala: Let's not use `loc` and `scale` here to keep this a truly Normal distribution and rather create an issue to add a TruncGaussian distribution that uses them.

copulas/univariate/truncnorm.py
```diff
@@ -38,10 +44,18 @@ def from_dict(cls, parameters):
...
 
-        if instance.fitted and instance.constant_value is None:
-            instance.model = scipy.stats.truncnorm(parameters['a'], parameters['b'])
+        if instance.fitted:
+            a = parameters['a']
+            b = parameters['b']
+            mean = parameters['mean']
+            std = parameters['std']
+
+            if a == b == mean and std == 0:
```
csala: Since we will be removing mean and std, let's encode this as `constant <=> a == b`

copulas/univariate/gaussian_kde.py
```diff
@@ -17,6 +17,25 @@ class GaussianKDE(ScipyWrapper):
     probability_density = 'evaluate'
     sample = 'resample'
 
+    def __init__(self, sample_size=None, *args, **kwargs):
+        self.sample_size = sample_size
```
csala: Let's add documentation about the `sample_size` in the class docstring

copulas/univariate/gaussian_kde.py
```diff
@@ -17,6 +17,25 @@ class GaussianKDE(ScipyWrapper):
...
+        super().__init__(*args, **kwargs)
+
+    def fit(self, X, *args, **kwargs):
+        self.constant_value = self._get_constant_value(X)
+
+        if self.constant_value is None:
+            if self.sample_size is not None:
+                model = scipy.stats.gaussian_kde(X)
```
csala: I would do `super.fit` instead of `scipy.stats.gaussian_kde`. Under the hoods is the same, but it is clearer that we are actually doing the same thing twice instead of doing two different things.

I would make the final implementation look like this:

```python
super().fit(X, *args, **kwargs)

if self.sample_size:
    X = self.sample(self.sample_size)
    super().fit(X, *args, **kwargs)
```

PR #103, opened 2019-08-27
Added support to python3.7

### LINKED ISSUE #53: Add support to Python 3.7
Currently, we only support python 3.5 and 3.6. We need to add the newest version of python. To do so, we need to check:
1. All dependencies of the package are compatible with `3.7`
2. The project builds after adding environment `3.7` on TravisCI
3. The supported versions are correctly listed in `setup.py`

JDTheRipperPC:
Resolves #53 

Updated tox.ini, setup.py and .travis.yml to support python3.7

If we don't support python versions higher that python 3.7, we must change `python_requires='>=3.5,` to `python_requires='>=3.5, <3.8`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 # Config file for automatic testing at travis-ci.org
+dist: xenial
 language: python
 python:
+  - 3.7
   - 3.6
   - 3.5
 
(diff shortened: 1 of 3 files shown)


csala:
Thanks @JDTheRipperPC !
No, I think it's OK to leave the version open on the upper side for now.

PR #138, opened 2020-02-12
Added gamma distribution

k15z:
This adds the gamma distribution to round out the set of 1d distributions that we are planning to support in the next release. It also moves the KS statistic inside the univariate module since we will be using a different and incomparable statistic for bivariate distributions.


csala:
@k15z would you mind creating an issue for this and adding the `Resolves #<number-of-issue>` in the description of the PR? Just to keep track of all the work done as issues.

Once it's created you can merge this.

k15z:
Resolves #141

PR #148, opened 2020-02-19
Transform inputs to `pdf`/`cdf` for gaussian multivariate

### LINKED ISSUE #147: Investigate `get_lower_bound` in GaussianMultivariate
The lower bound returned by this function is supposed to be the lower bound for the integral in the CDF. However, it's not clear what this is actually doing - in addition to the fact that many univariate distributions don't actually have a `mean` attribute, it's also not clear what the following line of code actually does:

> lower_bound = distribution.percent_point(distribution.mean / 10000)

Percent point is supposed to take in a probability and return the quantile, but here it's being given some random quantity that comes from the mean?

k15z:
Resolve #147 

The inputs to `pdf`/`cdf` are no longer expected to be the result of transforming X to uniform and then to gaussian. Instead, it directly takes X now. In addition, it also uses a more numerically stable implementation of `cdf` provided by scipy. Also, fixes a bug where off-diagonal covariances were being zero'd out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ clean-pyc: ## remove Python file artifacts
 .PHONY: clean-docs
 clean-docs: ## remove previously built docs
 	rm -f docs/api/*.rst
-	rm -f docs/tutorials/*
+	rm -rf docs/tutorials/*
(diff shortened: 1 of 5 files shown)


k15z:
Resolves #147

PR #152, opened 2020-02-24
Misc. bug fixes.

k15z:
Resolves XXX (todo). For context, during a manual review of the library, I identified the following potential issues.

 - Vectorized CDF for KDE estimator (significantly faster).
 - Alternative Beta distribution fit procedure, more robust + higher KS values.
 - Use the conditional PPF for bivariates (instead of re-estimating with unstable optimizer).
 - Derive the actual cumulative PDF for IndependenceCopula; I'm pretty sure it's wrong right now.
 - Switch to scipy's kstest implementation.

This draft PR shows how some of these can be addressed.


k15z:
For context, during a manual review of the library, I identified the following potential issues.

1. Vectorized CDF for KDE estimator (significantly faster).
2. Alternative Beta distribution fit procedure, more robust + higher KS values.
3. Use the conditional PPF for bivariates (instead of re-estimating with unstable optimizer).
4. Remove or explain the probability_density method on Vine/Tree - it's not clear what it actually does.
5. Remove tau from attributes / to_dict field. in bivariates
6. Remove u_matrix/tau_matrix from Vine attributes.
7. Derive the actual cumulative PDF for IndependenceCopula; I'm pretty sure it's wrong right now.
8. Switch to scipy's kstest implementation.

This draft PR shows how some of these can be addressed.

copulas/univariate/gaussian_kde.py
```diff
@@ -109,7 +107,7 @@ def percent_point(self, U):
         """
         self.check_fit()
 
-        return scipy.optimize.brentq(self._brentq_cdf(U), -1000.0, 1000.0)
```
k15z: What if you're data is outside of the range [-1000.0, 1000.0]? For example, timestamps?

copulas/univariate/selection.py
```diff
@@ -1,21 +1,15 @@
 import numpy as np
 
+from scipy.stats import kstest
 from copulas import get_instance
 
-
-def ks_statistic(model, X):
```
k15z: This was my bad, I didn't realize scipy had an implementation when I implemented, let's use it.

copulas/univariate/beta.py
```diff
@@ -36,7 +37,9 @@ def _get_model(self):
     def _fit_beta(self, X):
         """Fit the beta parameters to the data.
         """
-        self.a, self.b, self.loc, self.scale = beta.fit(X)
+        self.loc = np.min(X)
```
k15z: To be consistent with truncated gaussian, we should estimate loc/scale directly.

copulas/bivariate/independence.py
```diff
@@ -76,4 +76,4 @@ def percent_point(self, y, V):
 
         """
         self.check_fit()
-        return V
+        return y
```
k15z: I'm pretty sure this is supposed to be `y`, not `V`. The entire point of the independence copula is that U and V are independent, so why is the inverse conditional CDF dependent on V?

copulas/univariate/gaussian_kde.py
```diff
@@ -61,15 +64,10 @@ def cumulative_distribution(self, X):
...
-        self.check_fit()
-
-        low_bounds = self.model.dataset.mean() - (5 * self.model.dataset.std())
-
-        result = []
-        for value in X:
-            result.append(self.model.integrate_box_1d(low_bounds, value))
-
-        return np.array(result)
+        stdev = np.sqrt(self.model.covariance[0,0])
```
k15z: Let's vectorize this. It's much, much faster.

copulas/multivariate/vine.py
```diff
@@ -272,16 +272,13 @@ def _sample_row(self):
                         copula_type = current_tree[current_ind].name
                         copula = Bivariate(copula_type=CopulaTypes(copula_type))
                         copula.theta = current_tree[current_ind].theta
-                        derivative = copula.partial_derivative_scalar
```
k15z: This is re-implementing the conditional PPF using a numerical approximation instead of using the (potentially) closed form solution provided by the Copulas. Why is it doing this? I have no idea, but it's slower and less accurate.

copulas/bivariate/base.py
```diff
@@ -54,7 +54,7 @@ class Bivariate(object):
     theta_interval = []
     invalid_thetas = []
     theta = None
-    tau = None
+    tau = None  # TODO: Why is this an attribute?
```
k15z: Tau isn't a parameter. It's estimated once from the data and never used again. Why is it an attribute?

copulas/multivariate/vine.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, vine_type, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
         self.vine_type = vine_type
-        self.u_matrix = None
+        self.u_matrix = None # TODO: Decide whether this should be an attribute or not
```
k15z: Why is the univariate matrix stored as an attribute? First, I think that a "copula" shouldn't even be responsible transforming the data to produce univariate marginals. However, even if it responsible for it, why is it storing it as an instance attribute?

copulas/multivariate/vine.py
```diff
@@ -207,7 +207,10 @@ def train_vine(self, tree_type):
             LOGGER.debug('finish building tree: {0}'.format(k))
 
     def get_likelihood(self, uni_matrix):
-        """Compute likelihood of the vine."""
+        """Compute likelihood of the vine.
+
+        TODO: explain what this is supposed to do and make it work
```
k15z: What is this function doing? There's a test case for it but I have no idea what it does and why it does it. Apparently uni_matrix is supposed to be a 1d array?

PR #158, opened 2020-02-26
Documentation refresh + built-in datasets.

### LINKED ISSUE #145: Add dataset module
We should provide some "simulated" datasets (as well as some real datasets) for people to use. This could look something like:

```
from copulas.datasets import load_distributions
```

where

```
def load_distributions(seed=42):
    """
    This dataset contains 6 columns, each of which corresponds to a different
    univariate distribution:
        bernoulli - a Bernoulli distribution with p=0.3
        bimodal - a mixture of two Gaussians at 0.0 and 10.0 with stdev=1
        uniform - a uniform distribution in [-1.0, 3.0]
        normal - a normal distribution at 1.0 with stdev=1
        constant - a constant value
        exponential - an exponential distribution at 3.0 with rate 1.0
    """
    ...
```

These datasets should primarily be pandas DataFrames which can be directly passed to the multivariate distributions for modeling.

k15z:
Resolves #145. Add new Jupyter notebooks which demonstrate the functionality of the library on several built-in datasets.


copulas/univariate/base.py
```diff
@@ -184,6 +184,9 @@ def to_dict(self):
         if not self.fitted:
             return result
 
+        if isinstance(self, Univariate):
```
k15z: Temporary workaround to make saving/loading instances of the Univariate base class work. We need to fix it properly at some point, but for now, I want to focus on writing the tutorials.

PR #159, opened 2020-02-27
Added end-to-end tests

### LINKED ISSUE #160: Fix GaussianMultivariate save/load.
Currently, this doesn't work if either:

1. The univariate distributions are manually specified and/or involve filtering on parametric/bounds.
2. There are constant values and the `Univariate` base class is used.

k15z:
Resolves #160.


copulas/multivariate/gaussian.py
```diff
@@ -70,7 +71,11 @@ def _get_covariance(self, X):
 
         """
         result = self._transform_to_normal(X)
-        return pd.DataFrame(data=result).cov().values
+        covariance = pd.DataFrame(data=result).cov().values
```
k15z: This adds noise to the diagonal if the matrix has a large condition number, allowing us to avoid issues with singular matrices in the `pdf`/`cdf`. Alternatively, we could just check for constant columns and remove them before modeling.

copulas/multivariate/gaussian.py
```diff
@@ -160,16 +165,21 @@ def sample(self, num_rows=1):
         return pd.DataFrame(data=res)
 
     def to_dict(self):
-        distributions = {
+        distribs = {
             name: distribution.to_dict() for name, distribution in self.distribs.items()
         }
+        distribution = self.distribution
+        if isinstance(self.distribution, dict):
```
k15z: This allows us to save instances of this model where the univariate marginals are manually specified. On a related note, we need to check whether the Univariate instance can be saved/loaded while preserving the Parametric/Bounded flags.

copulas/univariate/base.py
```diff
@@ -180,14 +180,18 @@ def to_dict(self):
...
-            'fitted': self.fitted,
+            'fitted': self.fitted
         }
 
         if not self.fitted:
             return result
 
         if get_qualified_name(self) == get_qualified_name(Univariate):
-            result['instance_type'] = get_qualified_name(self._instance)
+            if self.constant_value is not None:
```
k15z: This modifies the univariate base class so it can be saved/loaded even if the value is constant.

Issue #172, opened 2020-05-25
Univariate Student T


gbonomib:
Add student t marginal

csala:
Closed via #173

PR #173, opened 2020-05-25
added tstud marginal + testing + docs

gbonomib:
hi guys

see below. i am wrapping a the t stud from scipy. unit and regression testing created and passing

solves Issue [#172](https://github.com/sdv-dev/Copulas/issues/172) by @gbonomib


copulas/univariate/student_t.py
```diff
@@ -0,0 +1,33 @@
...
+    def _fit_constant(self, X):
+        self._params = {
+            'df': 100,
+            'loc': np.unique(X)[0],
+            'scale': 0
+        }
+
+    def _fit(self, X):
+        self._params = {
+            'df': t.fit(X)[0],
```
csala, line 27: Here you are computing `fit` three times unnecessarily.
I would compute it once and then make the dictionary in a second step:

```python
        df, loc, scale = t.fit(X)
        self._params = {
            'df': df,
            'loc': loc,
            'scale': scale,
        }
```

copulas/univariate/student_t.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    PARAMETRIC = ParametricType.PARAMETRIC
+    BOUNDED = BoundedType.UNBOUNDED
+
+    MODEL_CLASS = t
+
+    def _fit_constant(self, X):
+        self._params = {
+            'df': 100,
```
csala, line 20: I think that this `100` value will not be consistent across different constant value.
For example, a vector which is not constant but very close to 0, like `[0.00001, 0, 0....]` gets a `df` value of `~2` while, a vector which is very close to 1, like `[1.00001, 1, 1, ...]` gets a `df` value of `~39`.

However, `scipy.stats.t` DOES handle constant arrays as input, so let's take advantage of it doing the following:
1. Remove this method completely from this class, so there is no `_fit_constant`.
2. Make `_is_constant` always return `False`.
3. Also overwrite `_check_constant_value` to always return `False`.
4. Adapt the tests accordingly.

```python
    def _is_constant(self):
        return False

    def _check_constant_value(self, *args, **kwargs):
        return False
```

This way, the our class will never handle constant values on its own and we will be always delegating to scipy.

tests/unit/univariate/test_t.py
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+    def test__fit(self):
+        distribution = StudentTUnivariate()
+
+        data = t.rvs(size=1000, df=3, loc=1, scale=1)
+        distribution._fit(data)
+
+        assert distribution._params == {
+            'df': t.fit(data)[0],
```
csala, line 29: Let's not rely on third party computations on the checks.
The expected values should be specified as literal values and validated using numpy's `assert_all_close`.

You can use other univariate unit tests, like  `test_gamma.py` or `test_beta.py`, as reference.

copulas/univariate/__init__.py
```diff
@@ -4,13 +4,15 @@
 from copulas.univariate.gaussian import GaussianUnivariate
 from copulas.univariate.gaussian_kde import GaussianKDE
 from copulas.univariate.truncated_gaussian import TruncatedGaussian
+from copulas.univariate.student_t import StudentTUnivariate
```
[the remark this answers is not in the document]
gbonomib: solved (does not show the change here but in the the rows should be swapped)

copulas/univariate/student_t.py
gbonomib: committed solution is failing a bunch of test which were previously ok. Could you have a look and check if the change are in line with what you meant? Thank you!

gbonomib:
> Many thanks for the contribution @gbonomib !
> 
> Please have a look at the comments that I added about some things that I think can be tweaked to make it perfect.

@csala thank you for the feedback, much appreciated

copulas/univariate/student_t.py
gbonomib: Nevermind I sorted it out and fixed all tests.
Removing `_fit constant` generated some issue in downstream tests. So now `_fit_constant` simply returns `_fit` and the every testing issues is circumvented whilst the handling of constant fitting is properly delegated to `scipy.stats.t` as recommended

tests/end-to-end/univariate/test_student_t.py
```diff
@@ -0,0 +1,117 @@
...
+        cdf2 = model2.cumulative_distribution(sampled_data)
+        assert np.all(np.isclose(cdf, cdf2, atol=0.01))
+
+    def test_to_dict_constant(self):
+        model = StudentTUnivariate()
+        model.fit(self.constant)
+
+        params = model.to_dict()
+
+        assert len(params) == 4
```
csala, line 94: This test should also be updated. The test assets should be made by values, even if we use a high tolerance, not by types. Thea reason being that type checking is a very broad test that does not truly guarantee anything.

Also, since in this case we have the `scale` value fully under control but the `df` might suffer small differences depending on system floating point issues, let's validate the `df` value fuzzily using numpy.

So, let's change this to:

```python
        df = params.pop('df')
        assert np.isclose(df, 152801, atol=1)

        assert params == {
            'type': 'copulas.univariate.student_t.StudentTUnivariate',
            'loc': 5,
            'scale': 0,
        }
```

tests/unit/univariate/test_student_t.py
```diff
@@ -0,0 +1,37 @@
...
+        }
+        for key, value in distribution._params.items():
+            np.testing.assert_allclose(value, expected[key], rtol=0.1)
+
+    def test__is_constant_true(self):
+        distribution = StudentTUnivariate()
+
+        distribution.fit(np.array([1, 1, 1, 1]))
+
+        assert not distribution._is_constant()
```
csala, line 30: Since now we will be also returning `True` if the distribution is truly constant, let's revert this to `assert distribution._is_constant()`.

gbonomib:
> I think we are almost ready @gbonomib !
> 
> I just added a couple of comments to handle the constant edge case a bit differently to ensure consistency and make sure the tests are right.
> 
> Would you mind reviewing them?
> Once this is done I think we are ready to go.

Just embedded your changes. Thank you @csala.
I'm closing this one. Please reopen if necessary.

gbonomib:
Just I realized that "close and comment" was not the right button

PR #186, opened 2020-08-18
Update 01_Introduction_to_Copulas.ipynb

nazar-ivantsiv:
Fix value range in comment


csala:
Good catch @nazar-ivantsiv ! Thanks for fixing it!

PR #201, opened 2020-12-22
Vectorize cdf, ppf

### LINKED ISSUE #200: Make gaussian_kde faster
The current implementation of the GaussianKDE is very slow, particularly for sampling.

I believe the culprit is the line below, since it calls [scipy's `ndtr`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ndtr.html) so many times:

```python3
uppers = np.vstack([ndtr((x - self._model.dataset) / stdev)[0] for x in X])
```

k15z:
Resolve #200. I'm on vacation from FB (happy holidays @csala!) so I thought I'd take a look at this - I've been wanting to make this faster forever, this PR makes gaussian kde ~8x faster.

Here's some code for benchmarking it:

```
import numpy as np
from time import time
from copulas.univariate.gaussian_kde import GaussianKDE

X_train = np.random.uniform(size=1000)
X_test = np.random.uniform(size=1000)

model = GaussianKDE()
model.fit(X_train)

start = time()
y_slow = model.percent_point(X_test)
print(time() - start)

start = time()
y_fast = model.percent_point_fast(X_test)
print(time() - start)

np.abs(y_fast - y_slow).max()
```

The vectorized version of the ppf takes 0.332 seconds while the current version takes 2.898 seconds.

### DIFF
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -1,7 +1,7 @@
 from functools import partial
 
 import numpy as np
-from scipy.optimize import brentq
+from scipy.optimize import brentq, newton
 from scipy.special import ndtr
 from scipy.stats import gaussian_kde
 
@@ -99,10 +99,11 @@ class GaussianKDE(ScipyModel):
             NotFittedError:
                 if the model is not fitted.
         """
+        X = np.array(X)
         self.check_fit()
         stdev = np.sqrt(self._model.covariance[0, 0])
         lower = ndtr((self._get_bounds()[0] - self._model.dataset) / stdev)[0]
-        uppers = np.vstack([ndtr((x - self._model.dataset) / stdev)[0] for x in X])
+        uppers = ndtr((X[:,None] - self._model.dataset) / stdev)
         return (uppers - lower).dot(self._model.weights)
 
     def _brentq_cdf(self, value):
@@ -177,6 +178,45 @@ class GaussianKDE(ScipyModel):
         X[is_one] = float("inf")
         X[is_zero] = float("-inf")
         X[is_valid] = brentq(self._brentq_cdf(U[is_valid]), lower, upper)
+
+        return X
+
+    def percent_point_fast(self, U):
+        """Compute the inverse cumulative distribution value for each point in U.
+
+        Arguments:
+            U (numpy.ndarray):
+                Values for which the cumulative distribution will be computed.
+                It must have shape (n, 1) and values must be in [0,1].
+
+        Returns:
+            numpy.ndarray:
+                Inverse cumulative distribution values for points in U.
+
+        Raises:
+            NotFittedError:
+                if the model is not fitted.
+        """
+        self.check_fit()
+
+        assert len(U.shape) == 1
+
+        if np.any(U > 1.0) or np.any(U < 0.0):
+            raise ValueError("Expected values in range [0.0, 1.0].")
+
+        is_one = U >= 1.0 - EPSILON
+        is_zero = U <= EPSILON
+        is_valid = np.logical_not(np.logical_or(is_zero, is_one))
+
+        lower, upper = self._get_bounds()
+
+        def _f(X):
+            return U[is_valid] - self.cumulative_distribution(X)
+
+        X = np.zeros(U.shape)
+        X[is_one] = float("inf")
+        X[is_zero] = float("-inf")
+        X[is_valid] = newton(_f, np.zeros(U[is_valid].shape) + (lower+upper)/2.0)
 
         return X
 
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -9,7 +9,7 @@ from copulas import EPSILON, scalarize, store_args
 from copulas.univariate.base import BoundedType, ParametricType, ScipyModel
 
 
-def _bisect(f, xmin, xmax, tol=1e-6, verbose=False):
+def _bisect(f, xmin, xmax, tol=1e-8, maxiter=50):
     a, b = xmin, xmax
 
     fa = f(a)
@@ -17,11 +17,13 @@ def _bisect(f, xmin, xmax, tol=1e-6, verbose=False):
     assert (fa <= 0.0).all()
     assert (fb >= 0.0).all()
 
-    while (b - a).max() > tol:
+    for _ in range(maxiter):
         c = (a + b) / 2.0  # proposal
         fc = f(c)
         a[fc <= 0] = c[fc <= 0]
         b[fc >= 0] = c[fc >= 0]
+        if (b - a).max() < tol:
+            break
 
     return (a + b) / 2.0
 
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -1,12 +1,140 @@
 from functools import partial
 
 import numpy as np
-from scipy.optimize import brentq, newton
+from scipy.optimize import brentq
 from scipy.special import ndtr
 from scipy.stats import gaussian_kde
 
 from copulas import EPSILON, scalarize, store_args
 from copulas.univariate.base import BoundedType, ParametricType, ScipyModel
+
+
+def _bisect(f, xmin, xmax, tol=1e-6, verbose=False):
+    a, b = xmin, xmax
+
+    fa = f(a)
+    fb = f(b)
+    assert (fa <= 0.0).all()
+    assert (fb >= 0.0).all()
+
+    while (b - a).max() > tol:
+        c = (a + b) / 2.0  # proposal
+        fc = f(c)
+        a[fc <= 0] = c[fc <= 0]
+        b[fc >= 0] = c[fc >= 0]
+
+    return (a + b) / 2.0
+
+
+def _chandrupatla(f, xmin, xmax, verbose=False,
+                  eps_m=None, eps_a=None,
+                  maxiter=50, return_iter=False, args=(),):
+    """
+    This is adapted from [1] which implements Chandrupatla's algorithm [2]
+    which starts from a bracketing interval and, conditionally, swaps between
+    bisection and inverse quadratic interpolation.
+
+    [1] https://github.com/scipy/scipy/issues/7242#issuecomment-290548427
+    [2] https://books.google.com/books?id=cC-8BAAAQBAJ&pg=PA95
+    """
+    # Initialization
+    a = xmax
+    b = xmin
+    fa = f(a, *args)
+    fb = f(b, *args)
+
+    # Make sure we know the size of the result
+    shape = np.shape(fa)
+    assert shape == np.shape(fb)
+
+    fc = fa
+    c = a
+
+    # Make sure we are bracketing a root in each case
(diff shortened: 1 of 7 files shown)


copulas/univariate/gaussian_kde.py
```diff
@@ -99,10 +99,11 @@ def cumulative_distribution(self, X):
             NotFittedError:
                 if the model is not fitted.
         """
+        X = np.array(X)
```
csala, line 102: I'd do this after the `check_fit` call.

copulas/univariate/gaussian_kde.py
```diff
@@ -180,6 +181,45 @@ def percent_point(self, U):
 
         return X
 
+    def percent_point_fast(self, U):
```
csala, line 184: Once this is working, the old one is not needed anymore. So let's replace the old one with the new implementation and also get rid of everything related to the old one.

copulas/univariate/gaussian_kde.py
```diff
@@ -180,6 +181,45 @@ def percent_point(self, U):
...
+            numpy.ndarray:
+                Inverse cumulative distribution values for points in U.
+
+        Raises:
+            NotFittedError:
+                if the model is not fitted.
+        """
+        self.check_fit()
+
+        assert len(U.shape) == 1
```
csala, line 202: Let's not use a plain `assert` here, which is hard to interpret by the users, but rather an `if` and a `raise ValueError` indicating that we only support 1d arrays.

copulas/univariate/gaussian_kde.py
```diff
@@ -180,6 +181,45 @@ def percent_point(self, U):
...
+
+        if np.any(U > 1.0) or np.any(U < 0.0):
+            raise ValueError("Expected values in range [0.0, 1.0].")
+
+        is_one = U >= 1.0 - EPSILON
+        is_zero = U <= EPSILON
+        is_valid = np.logical_not(np.logical_or(is_zero, is_one))
```
csala, line 209: Not a big deal, but I think that `is_valid = ~(is_zero | is_one)` would be more readable and slightly faster (though difference is probably negligible).

copulas/univariate/gaussian_kde.py
```diff
@@ -180,6 +181,45 @@ def percent_point(self, U):
...
+
+        lower, upper = self._get_bounds()
+
+        def _f(X):
+            return U[is_valid] - self.cumulative_distribution(X)
+
+        X = np.zeros(U.shape)
+        X[is_one] = float("inf")
+        X[is_zero] = float("-inf")
+        X[is_valid] = newton(_f, np.zeros(U[is_valid].shape) + (lower + upper) / 2.0)
```
csala, line 219: You could use `np.full` instead of doing `np.zeros + value`, right?

tests/large_scale_evaluation.py
```diff
@@ -37,14 +37,14 @@
 from datetime import datetime
 from urllib.parse import urljoin
 
-import boto3
 import numpy as np
 import pandas as pd
+from scipy.stats import ks_2samp
+
```
csala, line 43: This resorting of the imports does not seem necessary. I suspect that the environment which you are using to test this does not have `boto3`, `botocore` and `tabulate` installed, which is why `isort` thinks they are first party dependencies. They should all be installed when you run `make install-develop`.

csala (changes requested):
Hey! Thanks for the proposal @k15z !

The idea looks good, and I agree that vectorizing seems to be the way to go here, but we will need to work a bit more on it to make it robust, since the current implementation seems to fails in some edge cases.

Here's an example of such failure (when passing an array with a single element):

```python
In [1]: import numpy as np
   ...: from copulas.univariate import GaussianKDE
   ...: from copulas.datasets import sample_univariate_bimodal
   ...: 
   ...: data = sample_univariate_bimodal()
   ...: model = GaussianKDE()
   ...: model.fit(data)
   ...: model.percent_point_fast(np.array([0.5]))
   ...: 
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-2ed68b5e8edf> in <module>
      6 model = GaussianKDE()
      7 model.fit(data)
----> 8 model.percent_point_fast(np.array([0.5]))

~/Projects/MIT/Copulas/copulas/univariate/gaussian_kde.py in percent_point_fast(self, U)
    265         X[is_one] = float("inf")
    266         X[is_zero] = float("-inf")
--> 267         X[is_valid] = newton(_f, np.zeros(U[is_valid].shape) + (lower+upper)/2.0)
    268 
    269         return X

~/.virtualenvs/Copulas/lib/python3.8/site-packages/scipy/optimize/zeros.py in newton(func, x0, fprime, args, tol, maxiter, fprime2, x1, rtol, full_output, disp)
    338                             " Failed to converge after %d iterations, value is %s."
    339                             % (itr + 1, p1))
--> 340                         raise RuntimeError(msg)
    341                     warnings.warn(msg, RuntimeWarning)
    342                 p = (p1 + p0) / 2.0

RuntimeError: Tolerance of [-80.25251] reached. Failed to converge after 3 iterations, value is [34.01875655].
```

I also added some more comments in-line

copulas/univariate/gaussian_kde.py
k15z: yep

copulas/univariate/gaussian_kde.py
k15z: There are still some numerical stability issues, I'll remove it after they're resolved.

k15z:
Good point re: the stability issues. Updated to use bisect and chandrupatla, both of which are equal to or better than brentq in terms of robustness. Here are the running times:

```
Original: 30.22
Bisect: 5.47
Chandrupatla: 4.72
```

Up to you whether you prefer the simple bisect method or chandrupatla's algorithm... the latter is obviously slightly faster but is significantly more complex (and may potentially be less robust, although I'm not aware of any issues?)

copulas/univariate/gaussian_kde.py
```diff
@@ -9,6 +9,134 @@
 from copulas.univariate.base import BoundedType, ParametricType, ScipyModel
 
 
+def _bisect(f, xmin, xmax, tol=1e-6, verbose=False):
+    a, b = xmin, xmax
+
+    fa = f(a)
+    fb = f(b)
+    assert (fa <= 0.0).all()
```
k15z, line 17: Remove verbose, replace assert with user friendly error.

copulas/univariate/gaussian_kde.py
```diff
@@ -133,7 +262,7 @@ def f(x):
 
         return f
 
-    def percent_point(self, U):
+    def percent_point_slow(self, U):
```
k15z, line 265: Remove this after verifying the new method is good to go. Will also require removing helper methods like `_brentq_cdf`.

tests/unit/univariate/test_gaussian_kde.py
```diff
@@ -165,6 +166,19 @@ def test_percent_point_invalid_value(self):
         with self.assertRaises(ValueError):
             instance.percent_point(np.array([2.]))
 
+    def test_percent_point_convergence(self):
+        instance = GaussianKDE()
+        instance.fit(sample_univariate_bimodal())
+        x = instance.percent_point(np.array([0.5]))
+        assert abs(instance.cumulative_distribution(x) - 0.5) < 0.01
+
+    def test_percent_point_boundary_values(self):
```
k15z, line 175: Regression test, if all values are 0.0/1.0 (-inf/inf), it used to throw a random error.

csala:
> Good point re: the stability issues. Updated to use bisect and chandrupatla, both of which are equal to or better than brentq in terms of robustness (both are more stable the newton). Here are the running times:
> 
> ```
> Original: 30.22
> Bisect: 5.47
> Chandrupatla: 4.72
> ```
> 
> Up to you whether you prefer the simple bisect method or chandrupatla's algorithm... the latter is obviously slightly faster but is significantly more complex (and may potentially be less robust, although I'm not aware of any issues?)

I think that both `bisect` and `chandrupatla` options are good, but I'm getting slightly different times:

```python
In [1]: import numpy as np
   ...: from copulas.univariate import GaussianKDE
   ...: from copulas.datasets import sample_univariate_bimodal
   ...: 
   ...: data = sample_univariate_bimodal()
   ...: model = GaussianKDE()
   ...: model.fit(data)
   ...: cdf = model.cumulative_distribution(data)

In [2]: %timeit model.percent_point_slow(cdf[0:10])
34.3 ms ± 299 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit model.percent_point(cdf[0:10], method='bisect')
20 ms ± 83.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit model.percent_point(cdf[0:10], method='chandrupatla')
9.23 ms ± 212 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit model.percent_point_slow(cdf)
3.35 s ± 4.37 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit model.percent_point(cdf, method='bisect')
2.53 s ± 77.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit model.percent_point(cdf, method='chandrupatla')
1.15 s ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit model.percent_point_slow(np.concatenate([cdf] * 10))
33.8 s ± 97.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit model.percent_point(np.concatenate([cdf] * 10), method='bisect')
12.9 s ± 48.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [10]: %timeit model.percent_point(np.concatenate([cdf] * 10), method='chandrupatla')
5.84 s ± 18.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

It seems like `bisect` takes twice the time `chandrupatla` takes, and this is ~[1/3-1/6] of what the original approach took.

On the other side, it seems like both `bisect` and `chandrupatla` produce the same results as the original method, which is also be the right output (it successfully inverts the cdf):

```python
In [11]: np.allclose(data, model.percent_point_slow(cdf))
Out[11]: True

In [12]: np.allclose(data, model.percent_point(cdf, method='bisect'))
Out[12]: True

In [13]: np.allclose(data, model.percent_point(cdf, method='chandrupatla'))
Out[13]: True
```

Altogether, I think we could keep `chandrupatla` as the default option but also have the `bisect` as an alternative.

Before we call it done I would add a few changes:

1. Can we move the `bisect` and `chandruptla` functions to a `copulas.optimize` module?
2. Maybe we could simplify a bit the code and remove unused options, like the `return_iter` and also put longer variable names, etc. to improve readability.
3. We should add a few numerical tests to validate that both the outputs of the `chandrupatla` and `bisect` functions are right, and also numerical tests that validate that the `percent_point` is successfully inverting the `cumulative_distribution` (this could actually be added for all the distributions?)

csala:
Also, the build errors should be fixed after merging #203 to master and then merging master to this branch

csala (approved):

PR #203, opened 2021-01-04
Update github actions

fealho:
Updates the CI to use invoke (as opposed to make) to simplify running development and testing commands in Windows. It also adds support for all types of float/integer, instead of only float64/integer64.


csala (approved):
Looks good!

PR #216, opened 2021-03-06
Add matplotlib show() to quickstart

jakesyl:
Adds plot launch to quickstart

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ synthetic data that resembles it.
 import warnings
 warnings.filterwarnings('ignore')
 
+from matplotlib.pyplot import show
+
 from copulas.datasets import sample_trivariate_xyz
 from copulas.multivariate import GaussianMultivariate
 from copulas.visualization import compare_3d
@@ -108,6 +110,7 @@ synthetic_data = copula.sample(len(real_data))
 
 # Plot the real and the synthetic data to compare
 compare_3d(real_data, synthetic_data)
+show()
 ```
 
 The output will be a figure with two plots, showing what both the real and the synthetic


csala:
Thanks for the suggestion @jakesyl 

However, the change that you are proposing does not seem to be necessary, since the plot should be showing up anyway without that statement. You can see it working on the corresponding notebook tutorial: https://github.com/sdv-dev/Copulas/blob/master/tutorials/README%20Quickstart.ipynb

If that is not the case for you, would you mind opening an issue indicating the details of your environment so we can debug it?

jakesyl:
@csala to confirm, this happens when not in the notebook. Here's a before ([env](https://repl.it/join/onfptgob-jakesyl))

image

And here's an after:

image


Tbh, I'm not sure why the notebook works and this doesn't since I don't see `%matplotlib inline`/anything to indicate `show()` is called

csala:
> @csala to confirm, this happens when not in the notebook.

Oh, fair enough!

> Tbh, I'm not sure why the notebook works and this doesn't since I don't see `%matplotlib inline`/anything to indicate `show()` is called

I was puzzled by that too, but it seems like the reason is the `import matplotlib.pyplot as plt` inside `copulas/visualization.py`, which in fact acts as an alternative to the `%matplotlib inline` instruction: https://github.com/ipython/ipython/issues/12190#issuecomment-598992644

csala:
@jakesyl It seems like the change prevents the quickstart from running properly on Mac, possibly due to the blocking behavior of `plt.show`, so I think we need to change the approach. An option would be to add the `plt.show()` instruction inside the `visualization.py` module, at the end of each one of the functions that just plot without returning anything. Would you be able to test if that would solve the problem too?

jakesyl:
@csala absolutely! Testing this out now

jakesyl:
Wasn't able to get this working. I think it's fine, since it works in the jupyter notebook

PR #253, opened 2021-04-26
Return ax where possible for integration with matplotlib

vinhloc30796:


copulas/visualization.py
```diff
@@ -93,6 +93,8 @@ def compare_2d(real, synth, columns=None, figsize=None):
     ax = real.plot.scatter(x, y, color='blue', alpha=0.5, figsize=figsize)
     ax = synth.plot.scatter(x, y, ax=ax, color='orange', alpha=0.5, figsize=figsize)
     ax.legend(['Real', 'Synthetic'])
+    
+    return ax
```
noamsgl, line 97: I have a learning question. Why does this help integrate better with matplotlib?

vinhloc30796: @noamsgl I'm new to this too - I think it allows people to reuse the resulting [`matplotlib.axes.Axes`](https://matplotlib.org/stable/api/axes_api.html#the-axes-class) object and make more changes to it as they see fit.

PR #262, opened 2021-07-22
Optimize ScipyModel performance

### LINKED ISSUE #261: Use scipy stats models directly without creating instances
Current Copulas code always creates an instance of each scipy stats model before calling its methods.

Creating such instances takes some time, but after the instance is created calling one of its methods does not seem to have any performance advantage over calling the method directly on the model class passing the parameters as arguments, without creating any instance.

```python3
In [1]: from scipy.stats import norm

In [2]: %timeit norm(loc=100, scale=10).rvs(size=100)
484 µs ± 16.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit norm.rvs(loc=100, scale=10, size=100)
34.9 µs ± 1.39 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: instance = norm(loc=100, scale=10)

In [5]: %timeit instance.rvs(size=100)
35.2 µs ± 1.09 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

When a single instance is created once and then used multiple times to generate random values this does not seem to be an issue, but when multiple instances are being created with different parameters the instances creation can end up adding to a lot of unnecessarily lost time.

We can prevent this by changing the code to call the stats methods directly without creating and storing intermediate instances in the Univariate models.

csala:
Resolve #261 

This changes the way the scipy stats are used by calling their methods directly without creating unnecessary intermediate instances.

It also adds an empty `__init__` to the `ScipyModel` to skip the `store_args` decorator, which is not necessary here, and the candidate initialization being performed on the parent `Univariate.__init__`.

### DIFF
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -177,6 +177,7 @@ class GaussianKDE(ScipyModel):
 
     def _set_params(self, params):
         """Set the parameters of this univariate.
+
         Args:
             params (dict):
                 Parameters to recreate this instance.
--- a/copulas/univariate/gaussian_kde.py
+++ b/copulas/univariate/gaussian_kde.py
@@ -177,10 +177,13 @@ class GaussianKDE(ScipyModel):
 
     def _set_params(self, params):
         """Set the parameters of this univariate.
-
         Args:
             params (dict):
                 Parameters to recreate this instance.
         """
-        super()._set_params(params)
-        self._model = self._get_model()
+        self._params = params.copy()
+        if self._is_constant():
+            constant = self._extract_constant()
+            self._set_constant_value(constant)
+        else:
+            self._model = self._get_model()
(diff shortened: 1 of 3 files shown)


copulas/univariate/gaussian_kde.py
```diff
@@ -167,9 +167,23 @@ def _fit(self, X):
...
 
     def _is_constant(self):
         return len(np.unique(self._params['dataset'])) == 1
 
     def _extract_constant(self):
         return self._params['dataset'][0]
+
+    def _set_params(self, params):
+        """Set the parameters of this univariate.
+        Args:
```
katxiao, line 180: nit: should there be an extra empty line here?

katxiao (approved):
lgtm!

amontanez24 (approved):
This lgtm!

copulas/univariate/gaussian_kde.py
csala: Good catch!

PR #265, opened 2021-07-30
GitHub issue 264 exception optimization converged

### LINKED ISSUE #264: Exception: Optimization converged to parameters that are outside the range allowed by the distribution.
### Environment Details

* Copulas version: 0.5.0
* Python version: 3.8
* Operating System: Ubuntu 20.04

### Error Description

After `scipy` 1.7.0 release, some distributions fail to fit because now scipy validates the learned parameters.
When this happens during univariate selection, the entire process crashes.

### Steps to reproduce

```
>>> import numpy as np
>>> from copulas.univariate import Univariate
>>> 
>>> vector = np.concatenate([np.zeros(29), np.ones(21)])
>>> univariate = Univariate()
>>> univariate.fit(vector)
/mnt/nvme0n1p2/xals/.virtualenvs/Copulas.clean/lib/python3.8/site-packages/scipy/optimize/minpack.py:175: RuntimeWarning: The iteration is not making good progress, as measured by the 
  improvement from the last ten iterations.
  warnings.warn(msg, RuntimeWarning)
/mnt/nvme0n1p2/xals/.virtualenvs/Copulas.clean/lib/python3.8/site-packages/scipy/optimize/optimize.py:761: RuntimeWarning: invalid value encountered in subtract
  np.max(np.abs(fsim[0] - fsim[1:])) <= fatol):
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/xals/Projects/MIT/Copulas.clean/copulas/univariate/base.py", line 225, in fit
    self._instance = select_univariate(selection_sample, self.candidates)
  File "/home/xals/Projects/MIT/Copulas.clean/copulas/univariate/selection.py", line 25, in select_univariate
    instance.fit(X)
  File "/home/xals/Projects/MIT/Copulas.clean/copulas/univariate/base.py", line 599, in fit
    self._fit(X)
  File "/home/xals/Projects/MIT/Copulas.clean/copulas/univariate/beta.py", line 28, in _fit
    a, b = beta.fit(X, loc=loc, scale=scale)[0:2]
  File "/mnt/nvme0n1p2/xals/.virtualenvs/Copulas.clean/lib/python3.8/site-packages/scipy/stats/_continuous_distns.py", line 62, in wrapper
    return fun(self, *args, **kwds)
  File "/mnt/nvme0n1p2/xals/.virtualenvs/Copulas.clean/lib/python3.8/site-packages/scipy/stats/_continuous_distns.py", line 662, in fit
    return super().fit(data, *args, **kwds)
  File "/mnt/nvme0n1p2/xals/.virtualenvs/Copulas.clean/lib/python3.8/site-packages/scipy/stats/_distn_infrastructure.py", line 2505, in fit
    raise Exception("Optimization converged to parameters that are "
Exception: Optimization converged to parameters that are outside the range allowed by the distribution.
```

### Steps to fix

The selection code already catches `ValueError` exceptions raised by the candidates:

https://github.com/sdv-dev/Copulas/blob/64b34e613e231b19c99b16fc55ef70277d91f069/copulas/univariate/selection.py#L30

This `except` should be changed to catch plain `Exceptions` instead.

csala:
Resolve #264 

Capture `Exception` instead of `ValueError`. Also update the tests to cover this case and simplify previous code.

### DIFF
--- a/copulas/univariate/selection.py
+++ b/copulas/univariate/selection.py
@@ -27,7 +27,7 @@ def select_univariate(X, candidates):
             if ks < best_ks:
                 best_ks = ks
                 best_model = model
-        except ValueError:
+        except Exception:
             # Distribution not supported
             pass
 
--- a/tests/unit/univariate/test_selection.py
+++ b/tests/unit/univariate/test_selection.py
@@ -1,54 +1,91 @@
-from unittest import TestCase
+from unittest.mock import Mock, patch
 
 import numpy as np
 from scipy.stats import truncnorm
 
-from copulas.univariate import GaussianKDE, GaussianUnivariate, TruncatedGaussian
+from copulas.univariate import BetaUnivariate, GaussianKDE, GaussianUnivariate, TruncatedGaussian
 from copulas.univariate.selection import select_univariate
 
 
-class TestSelectUnivariate(TestCase):
+def test_select_univariate_bimodal():
+    """
+    Suppose the data follows a bimodal distribution. The model selector should be able to
+    figure out that the GaussianKDE is best.
+    """
+    mask = np.random.normal(size=1000) > 0.5
+    mode1 = np.random.normal(size=1000) * mask
+    mode2 = np.random.normal(size=1000, loc=10) * (1.0 - mask)
+    bimodal_data = mode1 + mode2
 
-    def setUp(self):
-        size = 1000
-        np.random.seed(42)
+    candidates = [GaussianKDE, GaussianUnivariate, TruncatedGaussian]
 
-        # Binary Data
-        self.binary_data = np.random.randint(0, 2, size=10000)
+    model = select_univariate(bimodal_data, candidates)
 
-        # Truncated Normal
-        a, b, loc, scale = -1.0, 0.5, 0.0, 1.0
-        self.truncated_data = truncnorm.rvs(a, b, loc=loc, scale=scale, size=10000)
+    assert isinstance(model, GaussianKDE)
 
-        # Mixture of Normals
-        mask = np.random.normal(size=size) > 0.5
-        mode1 = np.random.normal(size=size) * mask
-        mode2 = np.random.normal(size=size, loc=10) * (1.0 - mask)
-        self.bimodal_data = mode1 + mode2
 
-    def test_select_univariate(self):
-        """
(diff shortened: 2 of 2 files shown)


sarahmish (approved):
LGTM! I have one question though, is there a way to track the type of exception raised in the test cases? Ideally we should change a general `Exception` to something specific if/when `scipy` makes this change.

copulas/univariate/selection.py
```diff
@@ -27,7 +27,7 @@ def select_univariate(X, candidates):
             if ks < best_ks:
                 best_ks = ks
                 best_model = model
-        except ValueError:
+        except Exception:
```
katxiao, line 30: Do we want to swallow all exceptions? I'm wondering if this would cause us to miss other potential errors in this function.

Maybe we could match on the exception error string we expect and raise the exception in all other cases.

csala: Well, that was precisely the main point of the change :-)
Scipy made changes that now make some fit errors show up as plain `Exception`, so we cannot narrow it any more than that.
When it comes to filtering by the Exception message, it would be a very fragile approach, because we could end up having to search for more than one message. think about other "known" Exceptions that we have not seen yet, but also think about the possibility of having message variations in future `scipy` releases.

But most importantly: Do we really want to let other exceptions go? I think we do not. If there is anything that makes a `candidate` crash while other candidates can be fitted without issues, we do not want the selection process to crash because of that one. We rather want the selection to discard the candidate and move on with the ones that work.

If anything, we could add an `if` block at the end that raises an explicit error message if ALL the candidates failed, but I would rather open a separate issue and PR for that, because it would not really be related to the issue that is being fixed on this PR.

katxiao: I see. I guess my next question is, do we want to be aware of the errors being thrown from candidates, even if we do not raise them? If so, we could potentially log them

katxiao (approved):

copulas/univariate/selection.py
csala: Good point @katxiao More than logging I would use a `warning`, so the user has the ability to either add them to logs or ignore them if the thing becomes to verbose. I'll add that on the new issue specification.

PR #267, opened 2021-08-11
Reorganize workflows and add codecov

csala:
Reorganize the github workflows to have one job per workflow instead of all the jobs in a single one.
Thanks to this change, if one specific job fails for any reason and needs to be re-executed all the other jobs do not need to be run again.

Also, add coverage report to pytest and `codecov` action to upload coverage to codecov.io


.github/workflows/lint.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
```
amontanez24, line 15: I like that we don't need to run this on every version now!

amontanez24 (approved):
Just some small question, but this LGTM!

fealho (approved):
lgtm. Just a quick question, did you test this in a private fork?

PR #283, opened 2021-11-04
Adding support for Python 3.9

### LINKED ISSUE #282: Add support for Python 3.9
### Problem Description

Copulas should run on Python 3.9 and pass all the tests

amontanez24:
resolves #282

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,8 +13,8 @@ with open('HISTORY.md', encoding='utf-8') as history_file:
 
 install_requires = [
     'matplotlib>=3.2.0,<4',
-    'numpy>=1.18.0,<1.20.0;python_version<3.7',
-    'numpy>=1.20.0,<2;python_version>=3.7',
+    "numpy>=1.18.0,<1.20.0;python_version<'3.7'",
+    "numpy>=1.20.0,<2;python_version>='3.7'",
     'pandas>=1.1.3,<2',
     'scipy>=1.5.4,<2',
 ]
--- a/setup.py
+++ b/setup.py
@@ -13,9 +13,9 @@ with open('HISTORY.md', encoding='utf-8') as history_file:
 
 install_requires = [
     'matplotlib>=3.2.0,<4',
-    'numpy>=1.18.0,<2',
-    'pandas>=1.1,<2',
-    'scipy>=1.4.1,<2',
+    'numpy>=1.20.0,<2',
+    'pandas>=1.1.3,<2',
+    'scipy>=1.5.4,<2',
 ]
 
 development_requires = [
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -13,9 +13,10 @@
 
 install_requires = [
     'matplotlib>=3.2.0,<4',
-    'numpy>=1.18.0,<2',
-    'pandas>=1.1,<2',
-    'scipy>=1.4.1,<2',
+    "numpy>=1.18.0,<1.20.0;python_version<'3.7'",
```
katxiao, line 16: py3.6 still works with `numpy<2` right? Is there any reason we're capping it at `1.20.0`.

setup.py
```diff
@@ -62,7 +63,7 @@
 ]
 
 tutorials_require = [
-    'scikit-learn>=0.22,<0.23',
+    'scikit-learn>=0.24,<0.25',
```
katxiao, line 66: How did you determine this range?

setup.py
amontanez24: Good point. I think I originally did this to test for sure it was working. I can expand the range again

setup.py
amontanez24: I think `0.24` was the minimum for python 3.9. I just kept it at a one version gap

amontanez24: From https://pythonrepo.com/repo/scikit-learn-scikit-learn-python-deep-learning:
```
0.24.0(Dec 22, 2020)
We're happy to announce the 0.24 release. You can read the release highlights under https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_0_24_0.html and the long version of the change log under https://scikit-learn.org/stable/whats_new/v0.24.html#version-0-24-0

This version supports Python versions 3.6 to 3.9.
```

csala: What about 1.0 and above @amontanez24 ?

amontanez24: I haven't tried it. They are definitely Python 3.9 compatible, I just wasn't sure if it would cause anything to crash

csala: I think they mentioned that the 1.0 was not really a "big" change, so 1.0 was kind of equivalent to what a 0.25 would have been, and that they only changed it to 1.0 because they thought that the library was mature enough.
So I would say that if we extend it to `<2` and all the tests pass, we can go ahead with it.

setup.py
amontanez24: Sorry I got confused, python 3.6 doesn't work with anything after 1.20.0. That's the whole reason for this

amontanez24:
```
NumPy 1.20.0 Release Notes
This NumPy release is the largest so made to date, some 684 PRs
contributed by 184 people have been merged. See the list of highlights
below for more details. The Python versions supported for this release
are 3.7-3.9, support for Python 3.6 has been dropped. Highlights are

```

katxiao (approved):
lgtm

PR #288, opened 2021-11-10
Add working libraries

### LINKED ISSUE #80: Implement `partial_derivative_scalar` in Bivariate Base class
The method `_partial_derivative` is being used in the Vine classes:

https://github.com/DAI-Lab/Copulas/blob/57e4eb3a462e0ccffc25cc4bedd5a413304fe27a/copulas/multivariate/vine.py#L187

However, this method should is not intended to be called from outside the Bivariate copula classes (as it starts with an underscore), and is not implemented in all the Bivariate subclasses.

In order to fix this, the method should be moved to the Bivariate class and renamed to `partial_derivative_scalar` or similar.

### LINKED ISSUE #278: Increase code style lint
### Problem Description

Currently our code is being validated only by `flake8` 'vanilla' and just a few plugins. We would like to increase the code style checks by adding more add-on's that follow our code style and our standards.

Also we would like to ensure that our docstrings are properly written and follow the rest of our format.

### Additional context

We have performed this task already on [RDT](https://github.com/sdv-dev/RDT/) , more precisely on the following issue:
https://github.com/sdv-dev/RDT/issues/248#issuecomment-932085083

### Docstring plugin

We need to add `pydocstyle` plugin with the following lines on our `setup.cfg` file as we are following the [google convention](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings).

```
[pydocstyle]
convention = google
add-ignore = D107, D407, D417
```

### Flake8 plugins to be added
Flake8 comes with a lot of different addons that we can use to adapt it to our codestyle and checking, here is a list of plugins that I found to be interesting for us:

* `flake8-builtins` - Check for python builtins being used as variables or parameters.
* `flake8-comprehensions` - Helps you write better list/set/dict comprehensions.
* `flake8-debugger` - Debug statement checker.
* `flake8-variables-names` - Extension that helps to make more readable variables names.
* `Dlint` - Tool for encouraging best coding practices and helping ensure Python code is secure.
* `flake8-mock` - Provides checking mock non-existent methods.
* `flake8-fixme` - Check for FIXME, TODO and other temporary developer notes.
* `flake8-eradicate` - Plugin to find commented out or dead code.
* `flake8-mutable` - Extension for mutable default arguments.
* `flake8-print` - Check for print statements in python files.
* `flake8-pytest-style` - Plugin checking common style issues or inconsistencies.
* `flake8-quotes` - Extension for checking quotes in python.
* `flake8-multiline-containers` - Plugin to ensure a consistent format for multiline containers.
* `pandas-vet` - Plugin that provides opinionated linting for pandas code.
* `pep8-naming` - Check the PEP-8 naming conventions.
* `flake8-expression-complexity` - Plugin to validate expressions complexity.
*  `flake8-sfs` - String formatting.

fealho:
Resolve #278 .

Initially, this PR will add the addons which the code already passes.

This is the base PR in which all other addons will be added (each one will be implemented in a different PR).

### DIFF
--- a/copulas/visualization.py
+++ b/copulas/visualization.py
@@ -1,3 +1,5 @@
+"""Visualization utilities for the Copulas library."""
+
 import pandas as pd
 
 try:
@@ -68,6 +70,14 @@ def hist_1d(data, fig=None, title=None, position=None, bins=20, label=None):
 
 
 def side_by_side(plotting_func, arrays):
+    """Plot side-by-side figures.
+
+    Args:
+        plotting_func (callable):
+            A matplotlib function which takes in the standard plot kwargs.
+        arrays (dict[str, np.ndarray]):
+            A mapping from the name of the subplot to the values.
+    """
     fig = plt.figure(figsize=(10, 4))
 
     position_base = f'1{len(arrays)}'
@@ -79,6 +89,18 @@ def side_by_side(plotting_func, arrays):
 
 
 def compare_3d(real, synth, columns=None, figsize=(10, 4)):
+    """Generate a 3d scatter plot comparing real/synthetic data.
+
+    Args:
+        real (pd.DataFrame):
+            The real data.
+        synth (pd.DataFrame):
+            The synthetic data.
+        columns (list):
+            The name of the columns to plot.
+        figsize:
+            Figure size, passed to matplotlib.
+    """
     columns = columns or real.columns
     fig = plt.figure(figsize=figsize)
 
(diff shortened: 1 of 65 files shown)


copulas/visualization.py
```diff
@@ -111,7 +111,7 @@ def compare_1d(real, synth, columns=None, figsize=None):
     fig = plt.figure(figsize=figsize)
 
     for idx, column in enumerate(columns):
-        position = prefix + str(idx + 1)
+        position = int(prefix + str(idx + 1))
```
fealho, line 114: This change is needed to avoid the following error: `ValueError: Single argument to subplot must be a three-digit integer, not '211'`. This seems unrelated to my code changes, and due to an update in the API of matplotlib.

amontanez24 (approved):

pvk-developer (approved):
LGTM :shipit:

PR #290, opened 2021-11-10
Add `flake8-multiline-containers`

fealho:
As part of #278, add `flake8-multiline-containers` and update the code accordingly.

### DIFF
--- a/copulas/bivariate/__init__.py
+++ b/copulas/bivariate/__init__.py
@@ -153,10 +153,9 @@ def select_copula(X):
 
     empirical_aut = np.concatenate((empirical_left_aut, empirical_right_aut))
     candidate_auts = [
-        np.concatenate((left, right)) for left, right in zip(
-            candidate_left_auts, candidate_right_auts
-        )
-    ]  # noqa: JS102
+        np.concatenate((left, right))
+        for left, right in zip(candidate_left_auts, candidate_right_auts)
+    ]
 
     # compute L2 distance from empirical distribution
     diff_left = [np.sum((empirical_left_aut - left) ** 2) for left in candidate_left_auts]
--- a/copulas/bivariate/__init__.py
+++ b/copulas/bivariate/__init__.py
@@ -156,7 +156,7 @@ def select_copula(X):
         np.concatenate((left, right)) for left, right in zip(
             candidate_left_auts, candidate_right_auts
         )
-    ]
+    ]  # noqa: JS102
 
     # compute L2 distance from empirical distribution
     diff_left = [np.sum((empirical_left_aut - left) ** 2) for left in candidate_left_auts]
--- a/copulas/bivariate/base.py
+++ b/copulas/bivariate/base.py
@@ -220,9 +220,10 @@ class Bivariate(object):
     def generator(self, t):
         r"""Compute the generator function for Archimedian copulas.
 
-        The generator is a function :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`
-        that given an Archimedian copula fulills:
+        The generator is a function
+        :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`  # noqa: JS101
 
+        that given an Archimedian copula fulfills:
         .. math:: C(u,v) = \psi^{-1}(\psi(u) + \psi(v))
 
 
--- a/copulas/bivariate/clayton.py
+++ b/copulas/bivariate/clayton.py
@@ -14,9 +14,10 @@ class Clayton(Bivariate):
     def generator(self, t):
         r"""Compute the generator function for Clayton copula family.
 
-        The generator is a function :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`
-        that given an Archimedian copula fulills:
+        The generator is a function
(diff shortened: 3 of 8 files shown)


tests/unit/multivariate/test_tree.py
```diff
@@ -655,18 +655,18 @@ def test_get_likelihood_no_parents(self, bivariate_mock):
 
         expected_partial_derivative_call_args = [
             (
-                (np.array([[
+                (np.array([[  # noqa: JS101
```
fealho, line 658: @csala I'm not sure what to do about these. If I do follow the lint rule, I have to do the following:

```
(
  np.array(
    [
      [
        [0.25, 0.75],
        [0.50, 0.50],
      ]
    ]
  ),
), {}
```

which looks silly.

copulas/bivariate/base.py
```diff
@@ -220,9 +220,10 @@ def infer(self, X):
     def generator(self, t):
         r"""Compute the generator function for Archimedian copulas.
 
-        The generator is a function :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`
-        that given an Archimedian copula fulills:
+        The generator is a function
+        :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`  # noqa: JS101
```
fealho, line 224: The reason for this `noqa` is that `[0, \infty)` is interpreted as an opening `[` without closing bracket. I'm not sure how to avoid this misinterpretation, so I just `noqa`'d it.

copulas/bivariate/clayton.py
```diff
@@ -14,9 +14,10 @@ class Clayton(Bivariate):
     def generator(self, t):
         r"""Compute the generator function for Clayton copula family.
 
-        The generator is a function :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`
-        that given an Archimedian copula fulills:
+        The generator is a function
+        :math:`\psi: [0,1]\times\Theta \rightarrow [0, \infty)`  # noqa: JS101
```
fealho, line 18: Same as above.

copulas/bivariate/gumbel.py
```diff
@@ -25,9 +25,11 @@ def probability_density(self, X):
 
             \begin{align}
                 c(U,V)
-                    &= \frac{\partial^2 C(u,v)}{\partial v \partial u} \\
-                    &= \frac{C(u,v)}{uv} \frac{((-\ln u)^{\theta} + (-\ln v)^{\theta})^{\frac{2}
-                {\theta} - 2 }}{(\ln u \ln v)^{1 - \theta}} ( 1 + (\theta-1) \big((-\ln u)^\theta
+                    &= \frac{\partial^2 C(u,v)}{\partial v \partial u}
+                    &= \frac{C(u,v)}{uv} \frac{((-\ln u)^{\theta}  # noqa: JS101
```
fealho, line 29: These are multiline fractions. Since I couldn't find a way to write them in one line, I'm using `noqa`.

copulas/bivariate/__init__.py
```diff
@@ -156,7 +156,7 @@ def select_copula(X):
         np.concatenate((left, right)) for left, right in zip(
             candidate_left_auts, candidate_right_auts
         )
-    ]
+    ]  # noqa: JS102
```
fealho, line 159: I don't know why it complains about this one...

tests/unit/multivariate/test_tree.py
katxiao: Yeah it is a bit excessive, but my preference would be to do the formatting over ignoring the rule here.

copulas/bivariate/__init__.py
katxiao: That's strange. Would it help to add commas? e.g.

```
candidate_auts = [
    np.concatenate((left, right)) for left, right in zip(
        candidate_left_auts, candidate_right_auts,
    ),
]
```

csala: I think this is because once you add any multi-line in an expression, the plugin expects everything to be multi-line, and also the fact that it is a comprehension messes it a bit.

I suggest to change the splitting to right before the `for`, and then the plugin will not complain:

```python3
    candidate_auts = [
        np.concatenate((left, right))
        for left, right in zip(candidate_left_auts, candidate_right_auts)
    ]
```

tests/unit/multivariate/test_tree.py
csala: I agree, it's a bit excessive, and I also agree that in general it's better to re-format than to ignore the rule. It's also happening in only two places, so it's not that bad to have it fully expanded.

copulas/bivariate/__init__.py
fealho: @katxiao adding the comma doesn't work, I also thought of that 😀

@csala that works, thanks!

tests/unit/multivariate/test_tree.py
fealho: Done. I also created intermediate variables for them because the `flake8-expression-complexity` addon complains about these lines in a different PR.

pvk-developer (approved):

csala (approved):

PR #291, opened 2021-11-10
Add `flake8-expression-complexity`

fealho:
As part of #80, added `flake8-expression-complexity` addon and adapted our code to follow this style check.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -115,8 +115,8 @@ def tutorials(c):
 @task
 def lint(c):
     check_dependencies(c)
-    c.run('flake8 copulas --max-expression-complexity=8')
-    c.run('flake8 tests --ignore=D,SFS2 --max-expression-complexity=8')
+    c.run('flake8 copulas')
+    c.run('flake8 tests --ignore=D,SFS2')
     c.run('isort -c --recursive copulas tests')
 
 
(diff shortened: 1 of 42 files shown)


tasks.py
```diff
@@ -115,8 +115,8 @@ def tutorials(c):
 @task
 def lint(c):
     check_dependencies(c)
-    c.run('flake8 copulas')
-    c.run('flake8 tests --ignore=D,SFS2')
+    c.run('flake8 copulas --max-expression-complexity=8')
```
fealho, line 118: The standard limit is 7, but around 10 expressions have complexity of 7.5. Since it seemed silly to decrease the complexity of the expressions (they are just implementations of equations) I increased the cap.

amontanez24: Is there a way to put this in a config instead of in the command?

csala: @fealho can you point at some of these expressions?
In general the complexity can be decreased by capturing some things in intermediate variables, so I suspect we would want to do that.

fealho: Sure, I moved the limit back to 7 and I added the intermediate variables (you can check the specific examples in the code), I just think it doesn't necessarily improve code readability, so it's a silly change.

pvk-developer (approved):

csala (approved):

PR #292, opened 2021-11-10
Add `flake8-print`

fealho:
As part of #80, added `flake8-print` addon and adapted our code to follow this style check.

### DIFF
--- a/tests/large_scale_evaluation.py
+++ b/tests/large_scale_evaluation.py
@@ -234,7 +234,7 @@ def main():
 
     results = run_evaluation(model_names, dataset_names, args.max_rows, args.max_columns)
 
-    print(tabulate.tabulate(  # noqa: T001
+    print(tabulate.tabulate(
         results,
         tablefmt='github',
         headers=results.columns,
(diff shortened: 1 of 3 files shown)


tests/large_scale_evaluation.py
```diff
@@ -234,7 +234,7 @@ def main():
 
     results = run_evaluation(model_names, dataset_names, args.max_rows, args.max_columns)
 
-    print(tabulate.tabulate(
+    print(tabulate.tabulate(  # noqa: T001
```
fealho, line 237: I left the print as it seems innocuous, but we could remove it if we really don't want to allow prints.

amontanez24: This seems fine to me. @csala What do you think?

amontanez24 (approved):

tests/large_scale_evaluation.py
csala: It's fine, but rather than using the `noqa` here I would use `per-file-ignores`, since the ignore would be valid for the entire file (as it is a user script)

csala (approved):

PR #297, opened 2021-11-10
Add `pytest-style`

fealho:
As part of #278, added `pytest-style` addon and adapted our code to follow this style check.

### DIFF
--- a/tests/unit/bivariate/test_clayton.py
+++ b/tests/unit/bivariate/test_clayton.py
@@ -22,8 +22,8 @@ class TestClayton(TestCase):
     def test_fit(self):
         """On fit, theta and tau attributes are set."""
         self.copula.fit(self.X)
-        assert self.copula.tau, 0.7877
-        assert self.copula.theta, 7.4218
+        assert 0.7877 <= self.copula.tau < 0.7878
+        assert 7.4218 <= self.copula.theta < 7.4219
 
     def test_probability_density(self):
         """Probability_density returns the probability density for the given values."""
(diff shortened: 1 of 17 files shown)


katxiao (approved):
lgtm

tests/unit/bivariate/test_clayton.py
```diff
@@ -22,8 +22,8 @@ def setUp(self):
     def test_fit(self):
         """On fit, theta and tau attributes are set."""
         self.copula.fit(self.X)
-        self.assertAlmostEqual(self.copula.tau, 0.7877, places=3)
-        self.assertAlmostEqual(self.copula.theta, 7.4218, places=3)
+        assert self.copula.tau, 0.7877
```
csala, line 25: This change does not seem correct.
Here we are saying `assert self.copula.tau` and if it fails, give 0.7877 as the message.
I think we should rather be using [pytest.approx](https://docs.pytest.org/en/6.2.x/reference.html#pytest-approx)

tests/unit/bivariate/test_frank.py
```diff
@@ -23,8 +23,8 @@ def setUp(self):
     def test_fit(self):
         """On fit, theta and tau attributes are set."""
         self.copula.fit(self.X)
-        self.assertAlmostEqual(self.copula.tau, 0.7877, places=3)
-        self.assertAlmostEqual(self.copula.theta, 17.0227, places=3)
+        assert 0.7877 <= self.copula.tau < 0.7878
```
csala, line 26: All these could be done using `pytest.approx`, but this approach is good too!

tests/unit/bivariate/test_clayton.py
fealho: Oh, I forgot to write this test (that was the placeholder I left there), great catch! I'll use the same syntax I'm using in the other tests, for consistency (since you mentioned it's fine in the comment below).

csala (approved):

PR #298, opened 2021-11-10
Add `pep8-naming`

fealho:
As part of #278 , added `flake8-naming` addon and adapted our code to follow this style check.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,10 +34,20 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
-extend_ignore = W503,
-	N803,  # argument name 'X' should be lowercase 
-	N806,  # variable 'X' in function should be lowercase
-  SFS3  # String literal formatting using f-string.
+extend-ignore = D107,
+    D413,  # Missing docstring in __init__
+    DUO103, # insecure use of "pickle" or "cPickle"
+	  N803,  # argument name 'X' should be lowercase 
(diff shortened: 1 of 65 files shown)


tasks.py
```diff
@@ -115,8 +115,9 @@ def tutorials(c):
 @task
 def lint(c):
     check_dependencies(c)
-    c.run('flake8 copulas')
-    c.run('flake8 tests --ignore=D,SFS2')
+    c.run('flake8 copulas --ignore-names=X,U,V,W,A,B,T1,X_left_right,X_right_left,' \
```
fealho, line 118: @csala Not sure what we want to do on this one. Lint complains that they start with capital letters, but we since they are matrices, they should be capitalized.

csala: As discussed on call, the ignores are fine, but we should move them to `setup.cfg`

amontanez24 (approved):
Once @csala 's comment is addressed this looks good to go!

setup.cfg
```diff
@@ -34,8 +34,10 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
-ignore = W503
-extend-ignore = SFS3  # String literal formatting using f-string.
+extend_ignore = W503,
+	N803,  # argument name 'X' should be lowercase 
+	N806,  # variable 'X' in function should be lowercase
+  SFS3  # String literal formatting using f-string.
```
fealho, line 40: I'll fix the formatting the indentation here in the base PR, after this is merged.

csala (approved):

PR #300, opened 2021-11-10
Add `flake8-dlint`

fealho:
As part of #278, added `flake8-dlint` addon and adapted our code to follow this style check.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,7 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
+extend-ignore = DUO103 # insecure use of "pickle" or "cPickle"
 ignore = W503
 
 [isort]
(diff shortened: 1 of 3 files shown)


katxiao (approved):
lgtm

setup.cfg
```diff
@@ -34,6 +34,7 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
+extend-ignore = DUO103 # insecure use of "pickle" or "cPickle"
```
amontanez24, line 37: why is this error being raised and can we resolve it in the code?

fealho: I believe we want to ignore this error in general. @csala mentioned it [here](https://github.com/sdv-dev/SDMetrics/pull/96#discussion_r743669901).

csala: That is correct @fealho
We do not intend to stop using `pickle` for the load/save functionality, so we can only ignore this.

amontanez24 (approved):

PR #304, opened 2021-11-22
Add `flake8-variables-names`

fealho:
As part of #278, added `flake8-variables-names` addon. Rule VNE001 is being ignored for complaining about reasonable names like X and Y (discussed [here](https://github.com/sdv-dev/SDMetrics/pull/97#issuecomment-964123115)).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,6 +35,7 @@ universal = 1
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
 ignore = W503
+extend-ignore = VNE001 # Single letter variable names are not allowed
 
 [isort]
 line_length = 99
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -35,6 +35,7 @@ universal = 1
 max-line-length = 99
 exclude = docs, .git, __pycache__, .ipynb_checkpoints
 ignore = W503
+extend-ignore = VNE001 # Single letter variable names are not allowed
```
csala, line 38: I see there is an `ignore` above this `extend`.
I think we should have only one of them: either we add `VN001` to the `ignore` list above, or we move the `V503` ignore here. This second option may be also adding other ignores (the defaults) behind the curtains, though, so I think the first one is preferable.

pvk-developer (approved):

setup.cfg
fealho: Done. I planned to fix this in the base PR after everything was merged there, but I did it here instead.

amontanez24 (approved):

PR #313, opened 2022-02-04
Add method to set random seed on copulas models

### LINKED ISSUE #113: Use random_state instead of random_seed
The current random_seed is used to re-set the random_state every time the sample method is called.
As a result of this, the output of the sample method is always the same.

Instead of this, the `random_seed` argument should be renamed to `random_state` and allow both integers and `RandomState` instances as input.

Inside the `random_state` wrapper, the behavior should be as follows:
1. Capture the current random state and store it in a variable
2. If `self.random_state` is an integer, create a `RandomState` from it.
3. set `self.random_state` as the current numpy random state.
4. Execute the function within a try/finally block
5. get the current random_state and store it in `self.random_state`
6. restore the original numpy random state

katxiao:
As part of https://github.com/sdv-dev/SDV/issues/690, we want to add a fixed seed when sampling. This requires setting the seed at sample time (after the model has already been created). 

Currently, we can only pass in the `random_seed` at initialization of the model. In this PR, I add a setter method for `random_seed`, to enable to following flow:

1. create model
2. set seed
3. sample 
4. repeat steps 2-3 as needed.

Resolves #113


csala (commented):
This looks good @katxiao ! I want to have another look before approving, but the changes seem to be right on spot so far.

One note is: maybe we should consider also addressing issue #113 on this PR. This would basically fix the current `random_state` wrapper, which currently sets the `numpy` seed globally instead of just within the current operation, but also allow setting the random seed once and calling `sample` multiple times obtaining different results at each call (Notice that this would be necessary in order to allow `reject_sampling` strategies with a fixed seed!)

amontanez24 (approved):
This looks good to me, but I think @csala has a good point. Maybe we can add that change too

copulas/__init__.py
```diff
@@ -21,19 +21,29 @@ class NotFittedError(Exception):
...
+        desired_state = random_state
+
+    np.random.set_state(desired_state.get_state())
+
     try:
         yield
     finally:
-        np.random.set_state(state)
+        set_model_random_state(desired_state)
+        np.random.set_state(original_state)
```
amontanez24, line 46: does this need to be done no matter what, or only when an int is passed?

amontanez24 (approved):
I think this looks good!

csala:
> I'm a little confused, because even if we switch to setting the numpy state instead of the seed, aren't we still setting it globally?

Sorry, my original statement was not precise enough. It is true that every time we set the random state we do it globally, but what I actually meant was that the change was permanent, meaning that operations that come after ours would also be affected by that change. But this was actually wrong, because the original `random_seed` function already had the `try/finally` block that restored the original state after the operation is finished.

In any case, the real advantage of using the `random_state` instead of the `random_seed` is the other part: now this is a `state` which changes over time as calls happen, rather than a fixed `seed` that is always the same on every call.

copulas/__init__.py
```diff
@@ -21,19 +21,29 @@ class NotFittedError(Exception):
...
+        random_state (int or np.random.RandomState):
+            The random seed or RandomState.
+        set_model_random_state (function):
+            Function to set the random state on the model.
     """
-    state = np.random.get_state()
-    np.random.seed(seed)
+    original_state = np.random.get_state()
+
+    if isinstance(random_state, int):
```
csala, line 35: This `if` block should have 3 branches: `int`, `RandomState` and `tuple`, since passing a state like the one returned by `np.random.get_state()` should also be possible (see comment below)

```python3
if `int`:
    state = RandomState(state).get_state()
elif `RandomState`:
    state = state.get_state()
elif not `tuple`:
    raise TypeError()
```

copulas/__init__.py
```diff
@@ -21,19 +21,29 @@ class NotFittedError(Exception):
...
+    else:
+        desired_state = random_state
+
+    np.random.set_state(desired_state.get_state())
+
     try:
         yield
     finally:
-        np.random.set_state(state)
+        set_model_random_state(desired_state)
```
csala, line 45: We should not be caputring the `desired_state`, but rather the `current state of numpy` as returned by `np.random.get_state()`

csala (changes requested):
I think that there are a few changes required to achieve the desired functionality.

Here is an `ipython` snippet that explains the expected behavior a bit more precisely:

```python3
In [1]: import numpy as np

In [2]: # Background: We simulate an external seed of 42, which we
   ...: # do not want to alter, and we set our model seed to 0.
   ...: # For reference, these are the sequences of random numbers
   ...: # that each seed produces:
   ...: 
   ...: np.random.seed(42)
   ...: np.random.random(size=4)
Out[2]: array([0.37454012, 0.95071431, 0.73199394, 0.59865848])

In [3]: np.random.seed(0)
   ...: np.random.random(size=4)
Out[3]: array([0.5488135 , 0.71518937, 0.60276338, 0.54488318])

In [4]: # EXTERNAL: We simulate an external seed that we do not want to alter
   ...: # and certify that the random numbers are the expected ones
   ...: np.random.seed(42)
   ...: np.random.random(size=2)
Out[4]: array([0.37454012, 0.95071431])

In [5]: # FIRST CALL: Inside our decorator we capture the original state
   ...: # and set a new one
   ...: original_state = np.random.get_state()
   ...: new_state = np.random.RandomState(seed=0).get_state()
   ...: np.random.set_state(new_state)
   ...: 
   ...: # Certify the random numbers are the expected ones
   ...: np.random.random(size=2)
Out[5]: array([0.5488135 , 0.71518937])

In [6]: # We capture the state AFTER the call and restore the original one
   ...: post_state = np.random.get_state()
   ...: np.random.set_state(original_state)
   ...: 
   ...: # Certify that the original state is restored and the random
   ...: # sequence can continue as expected (sequence continues)
   ...: np.random.random(size=2)
Out[6]: array([0.73199394, 0.59865848])

In [7]: # SECOND CALL: Inside the decorator again, we restore the previous state
   ...: np.random.set_state(post_state)
   ...: 
   ...: # We certify that the sequence of random numbers with seed = 0 continues
   ...: # as expected
   ...: np.random.random(size=2)
Out[7]: array([0.60276338, 0.54488318])
```

Additionally, we should create an integration test that reproduces a sequence similar to the one
shown above and that certifies that multiple calls after setting the seed produce different results, but
always following the expected sequence.

katxiao:
> ```python
> np.random.random(size=2)
> ```

@csala I added the integration test and addressed the other two comments!

copulas/__init__.py
```diff
@@ -21,19 +21,32 @@ class NotFittedError(Exception):
...
+
+    if isinstance(random_state, int):
+        desired_state = np.random.RandomState(seed=random_state).get_state()
+    elif isinstance(random_state, np.random.RandomState):
+        desired_state = random_state.get_state()
+    elif not isinstance(random_state, tuple):
+        raise TypeError(f'RandomState {random_state} is an unexpected type. '
+                        'Expected to be int, np.random.RandomState, or tuple.')
+
+    np.random.set_state(desired_state)
```
csala, line 43: Bug: If `random_state` is a `tuple`, `desired_state` is never assigned any value. I think that it would be simpler to just re-use the `random_state` variable name instead of `desired_state`

csala (changes requested):
I added a comment about an edge case bug. Other than that, this looks ready

csala (commented):
I think this looks all correct so far. The only comment that I would add here is that it would be interesting to add one integration test per model, which certifies that the `set_random_state` is working as expected.

The tests I'm thinking about would be something like this (for each model!):

```python3
# Fit the model on some random data
fit_data = np.random.whatever...
model = Model()
model.fit(fit_data)

# Sample truly random data
random = model.sample(10)

# Set the seed to a fixed value and sample TWICE
model.set_random_seed(0)
seeded_0_0 = model,sample(10)
seeded_0_1 = model,sample(10)

# Set the seed again to the same value and sample TWICE again
model.set_random_seed(0)
seeded_1_0 = model,sample(10)
seeded_1_1 = model,sample(10)

# assert that the random data is not equal to the data with fixed seed
np.testing.assert_not_equals(random, sampled_0_0)
# assert that the two sample calls after setting the seed generated different outputs
np.testing.assert_not_equals(sampled_0_0, sampled_0_1)
# assert that setting the seed sampling once always produces the same results
np.testing.assert_equals(sampled_0_0, sampled_1_0)
# assert that the second call after setting the seed continues to produce the same results
np.testing.assert_equals(sampled_0_1, sampled_1_1)
```

tests/end-to-end/univariate/test_base.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
+import pytest
+from scipy.stats import norm
+
+from copulas.univariate import Univariate
+
+
+class TestUnivariate():
+
+    @pytest.mark.parametrize('model', Univariate._get_subclasses())
```
csala, line 10: In other cases parameterizing something like this would make sense.
However, since in this case each model has its own tests module [here](https://github.com/sdv-dev/Copulas/tree/master/tests/end-to-end/univariate) and all of them follow a similar structure, I think it would should add the test within each module, even if the function ends up being identical for each model.

If we consider that there is too much code duplication in these tests, we can open an issue to evaluate the possibility of unifying all the tests in a single tests module in the future, with just a different `setUp` for each one of them.

tests/end-to-end/multivariate/test_base.py
```diff
@@ -0,0 +1,38 @@
...
+    @pytest.mark.parametrize('model', [GaussianMultivariate, VineCopula])
+    def test_fixed_random_state(self, model):
+        """Test that the multivariate models work with a fixed seed.
+
+        Expect that fixing the seed generates a reproducable sequence
+        of samples. Expect that these samples are different from randomly
+        sampled results.
+        """
+        data = sample_trivariate_xyz()
+        if model == VineCopula:
```
csala, line 19: We can skip this `if/else` block by parameterizing with instances instead of classes:

```python
@pytest.mark.parametrize('model', [GaussianMultivariate(), VineCopula('direct')])
```

copulas/univariate/base.py
```diff
@@ -81,11 +81,28 @@ def _select_candidates(cls, parametric=None, bounded=None):
 
         return candidates
 
+    @classmethod
+    def _get_subclasses(cls):
```
csala, line 85: I do not think this method is needed, since it is equivalent to calling the `_select_candidates` method above without any arguments.

tests/end-to-end/bivariate/test_base.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+import pytest
+
+from copulas.bivariate import Bivariate
+
+
+class TestBivariate():
+
+    @pytest.mark.parametrize('model', Bivariate._get_subclasses())
```
csala, line 9: We should use the `subclasses()` method instead of `_get_subclasses`, since this one is intended to be used privately only, and they return the same information.

csala (changes requested):

csala (approved):
Looks good now!

PR #319, opened 2022-04-29
Return loc and scale from the fitted distribution

### LINKED ISSUE #317: Improve the fit of the Beta distribution: Use the new `loc` and `scale`
### Problem Description

In the [beta univariate fit function](https://github.com/sdv-dev/Copulas/blob/master/copulas/univariate/beta.py#L30), we perform the following steps:

1. Estimate the `loc` and `scale` parameters
2. Call the scipy `fit` function using the `loc` and `scale` as starting guesses
3. After the fit is complete, get the values for `a` and `b`

The issue is that step 3 also returns new `loc` and `scale` parameters. The ones we input are just starting guesses. When we use the same `loc` and `scale` as step 1, they are out-of-sync with the `a` and `b` parameters.

### Expected behavior
Stop using the initial guesses for `loc` and `scale`. Update them when setting `a` and `b`. 

i.e. change [line 30](https://github.com/sdv-dev/Copulas/blob/master/copulas/univariate/beta.py#L30)

```python
def _fit(self, X):
        loc = np.min(X)
        scale = np.max(X) - loc
        a, b, loc, scale = beta.fit(X, loc=loc, scale=scale)
        self._params = {
            'loc': loc,
            'scale': scale,
            'a': a,
            'b': b
        }
```

### Additional context

We verified this change by comparing our fit distribution to scipy. Scipy's fit is better because it's actually updating the `loc` and `scale` parameters.

Sometimes it's off by a little bit -- 
image

Sometimes, by a lot --
image

pvk-developer:
Resolve #317

### DIFF
--- a/copulas/univariate/beta.py
+++ b/copulas/univariate/beta.py
@@ -27,7 +27,7 @@ class BetaUnivariate(ScipyModel):
     def _fit(self, X):
         loc = np.min(X)
         scale = np.max(X) - loc
-        a, b = beta.fit(X, loc=loc, scale=scale)[0:2]
+        a, b, loc, scale = beta.fit(X, loc=loc, scale=scale)
         self._params = {
             'loc': loc,
             'scale': scale,
--- a/tests/unit/univariate/test_beta.py
+++ b/tests/unit/univariate/test_beta.py
@@ -1,4 +1,5 @@
 from unittest import TestCase
+from unittest.mock import patch
 
 import numpy as np
 from scipy.stats import beta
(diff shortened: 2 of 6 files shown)


copulas/univariate/beta.py
```diff
@@ -27,7 +27,7 @@ def _fit_constant(self, X):
     def _fit(self, X):
         loc = np.min(X)
         scale = np.max(X) - loc
-        a, b = beta.fit(X, loc=loc, scale=scale)[0:2]
+        a, b, loc, scale = beta.fit(X, loc=loc, scale=scale)
```
amontanez24, line 30: Can we add a small unit test to capture this change?

fealho (approved):
Looks good 👍

tests/unit/univariate/test_beta.py
```diff
@@ -61,3 +62,34 @@ def test__extract_constant(self):
...
+
+        Side Effect:
+            - The distribution has to learn the return values from the ``beta`` fit.
+        """
+        # Setup
+        distribution = BetaUnivariate()
+        mock_beta.fit.return_value = (1, 2, 3, 4)
+
+        # Run
+        distribution.fit(np.array([1, 2, 3, 4]))
```
fealho, line 87: I think you should test `_fit` directly, since this is a unit test.

PR #337, opened 2022-12-20
Package Maintenance Update

### LINKED ISSUE #336: Copulas Package Maintenance Updates
_This is a placeholder issue for overall package maintenance updates for the next minor Copulas release -- `0.8.0`_

pvk-developer:
Resolves #336


docs/conf.py
```diff
@@ -59,8 +59,8 @@
 project = 'Copulas'
 slug = 'copulas'
 title = project + ' Documentation',
-copyright = '2018, MIT Data To AI Lab'
-author = 'MIT Data To AI Lab'
+copyright = '2022, DataCebo Inc.'
+author = 'DataCebo, Inc.'
 description = 'A library to model multivariate data using copulas.'
```
npatki, line 64: I think we can use the same description as we use for the `setup.py` file --
```
'Create tabular synthetic data using copulas-based modeling'
```

npatki (approved):
LGTM! Left one suggestion for updating the description in the docs

setup.py
```diff
@@ -87,7 +87,7 @@
 ]
 
 tutorials_require = [
-    'scikit-learn>=0.24,<2',
+    'scikit-learn>=0.24,<1.2',
```
amontanez24, line 90: any reason we changed this?

amontanez24 (commented):
One question, but after that should be good!

setup.py
pvk-developer: Yes, the tutorials are not working since the demo datasets used are available until 1.2
Since this was for tutorials only I think it was okay to just cap it right away.

amontanez24 (approved):
:shipit:

PR #338, opened 2022-12-22
Support python 3.10 and above

pvk-developer:


setup.py
```diff
@@ -132,7 +131,7 @@
     keywords='copulas',
     name='copulas',
     packages=find_packages(include=['copulas', 'copulas.*']),
-    python_requires='>=3.6,<3.11',
+    python_requires='>=3.7,<4',
```
katxiao, line 134: Maybe we should cap this at `<3.12` since we don't support >= 3.11 yet.

katxiao (commented):
Looks good in general, just had one comment about capping the Python version

setup.py
```diff
@@ -12,16 +12,14 @@
     history = history_file.read()
 
 install_requires = [
-    "matplotlib>=3.2.0,<3.4;python_version<'3.7'",
     "matplotlib>=3.4.0,<4;python_version>='3.7' and python_version<'3.10'",
     "matplotlib>=3.6.0,<4;python_version>='3.10'",
-    "numpy>=1.18.0,<1.20.0;python_version<'3.7'",
     "numpy>=1.20.0,<2;python_version>='3.7' and python_version<'3.10'",
```
amontanez24, line 17: should we drop support for numpy<1.22 because of these security warnings? https://github.com/sdv-dev/Copulas/security/dependabot

Makefile
```diff
@@ -102,35 +102,28 @@ fix-lint: ## fix lint issues using autoflake, autopep8, and isort
 
 # TEST TARGETS
 
-.PHONY: test-unit
-test-unit: ## run unit tests using pytest
-	python -m pytest --disable-warnings --cov=copulas tests/unit
+.PHONY: test-end-to-end
+test-end-to-end: ## run tests quickly with the default Python
```
amontanez24, line 106: can we update the comment to say what it actually does? Is it running integration tests only?

tasks.py
```diff
@@ -19,6 +20,11 @@
     '==': operator.eq,
 }
 
+
+if not hasattr(inspect, 'getargspec'):
+    inspect.getargspec = inspect.getfullargspec
```
amontanez24, line 25: is this being tested on windows in any of the workflows? (ie. do any of them run `invoke`)

amontanez24 (commented):
This looks good. Just a few comments :)

tasks.py
pvk-developer: Yes, actually the workflows are being run using `invoke <task-name>` And if they are not this way we probably have to change it to be like this. There are some exceptions such as make docs and so on but other than that should be run using `invoke` (and the current workflows do).

setup.py
pvk-developer: Last supported version for numpy and python 3.7 is `1.21.6` 😥

amontanez24 (approved):

katxiao (approved):

katxiao:
One more thing - could you also update the supported Python versions in the [INSTALL.md](https://github.com/sdv-dev/Copulas/blob/master/INSTALL.md)?

PR #345, opened 2023-03-30
Add version add-on discovery

### LINKED ISSUE #344: Add functionality to find version add-on
### Problem Description

_As an enterprise user, it would be nice to see the version of public Copulas that is being used as well as the enterprise version._

The actual functionality for this will be added in SDV enterprise, however, this library will need to know to add that functionality to the proper name space.

### Expected behavior

- In this[ file](https://github.com/sdv-dev/Copulas/blob/master/copulas/__init__.py), add and call a function that uses an entry point called `copulas_module` to add the `version` module to the namespace.
- Add any necessary logging and error checking
- Be careful not to override anything in the globals that is needed

### Additional context
The code can look something like the following:

```python
def _add_versioning():
    for entry_point in pkg_resources.iter_entry_points('copulas_module'):
        try:
            version_module = entry_point.load()
        except Exception:
            message = f'Failed to load "{entry_point.name}" from "{entry_point.module_name}". '
            continue

        if 'version' not in globals():
            globals()['version'] = version_module

_add_versioning()
```

In this case, the loaded entry point should return the module itself, so we can just add it directly to the `globals()` under the name `version`. This should enable the following:
```python
>>> copulas.version.public
0.8.0
```

frances-h:
Resolve #344

### DIFF
--- a/copulas/__init__.py
+++ b/copulas/__init__.py
@@ -9,6 +9,9 @@ __version__ = '0.8.1.dev0'
 import contextlib
 import importlib
 from copy import deepcopy
+import warnings
+from pkg_resources import iter_entry_points
+
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


fealho (commented):
Same comments as I left in the SDV PR.

fealho (approved):

PR #350, opened 2023-04-21
Remove upper bound for pandas library

### LINKED ISSUE #349: Remove upper bound for pandas
### Problem Description

With the newly released [pandas 2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html) it would nice to support it across the sdv-dev libraries.
 
### Expected behavior

The newly expected version for pandas should be only `>=` than the minimum supported without upper bound. After the changes are done, both minimum and newest version should work properly.

pvk-developer:
Resolves #349

### DIFF
--- a/tests/unit/multivariate/test_gaussian.py
+++ b/tests/unit/multivariate/test_gaussian.py
@@ -423,7 +423,7 @@ class TestGaussianMultivariate(TestCase):
         result = instance.sample(5)
 
         # Check
-        pd.testing.assert_frame_equal(result, expected_result, check_less_precise=True)
+        pd.testing.assert_frame_equal(result, expected_result)
 
     def test_to_dict(self):
         """To_dict returns the parameters to replicate the copula."""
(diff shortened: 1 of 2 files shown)


tests/unit/multivariate/test_gaussian.py
```diff
@@ -423,7 +423,7 @@ def test_sample_random_state(self):
         result = instance.sample(5)
 
         # Check
-        pd.testing.assert_frame_equal(result, expected_result, check_less_precise=True)
```
amontanez24, line 426: did we need the `check_less_precise=True`?

pvk-developer: This is now deprecated, but I can do `check_exact=True` if we need to ensure exact numbers

fealho (approved):

tests/unit/multivariate/test_gaussian.py
pvk-developer: Actually they suggest using `rtol` 🤔

amontanez24: I don't think we were trying to check exact before but actually the opposite. The parameter was giving some tolerance and I'm wondering if not doing it anymore will lead to a fragile test. Here is another possible replacement https://github.com/pandas-dev/pandas/pull/30562

amontanez24: But if the test passes without any issues we can just get rid of it

PR #407, opened 2024-05-01
Add ruff for linting, remove flake8, remove isort, remove pylint

gsheni:
- Added [ruff](https://github.com/charliermarsh/ruff) for running linting
- Added **ruff** commands to Makefile
- Removed **flake8** as dev dependency, since **ruff** does this, but faster
- Remove **isort** as dev dependency, since **ruff** has the [same function](https://github.com/charliermarsh/ruff#how-does-ruffs-import-sorting-compare-to-isort), but faster
- Removed **pylint** as dev dependency, since **ruff** has [more implemented rules](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint)
- Removed setup.cfg
- Closes #88


pyproject.toml
```diff
@@ -199,3 +171,51 @@ filename = 'copulas/__init__.py'
 search = "__version__ = '{current_version}'"
 replace = "__version__ = '{new_version}'"
 
+[tool.ruff]
+preview = true
+line-length = 100
+indent-width = 4
+src = ["rdt"]
```
amontanez24, line 178: copulas?

tests/unit/multivariate/test_gaussian.py
```diff
@@ -3,41 +3,94 @@
...
+                1959.28188858,
+                1076.99295365,
+                2029.25100261,
+                1835.52188141,
+                1170.03850556,
+                739.42628394,
+                1866.65810627,
+                3703.49786503,
+                1719.45232017,
+                258.90206528,
```
amontanez24, line 32: I hate this

Makefile
```diff
@@ -82,23 +81,17 @@ install-test: clean-build clean-pyc ## install the package and test dependencies
...
 .PHONY: lint
-lint: ## check style with flake8 and isort
-	invoke lint
-
-lint-docs: ## check docs formatting with doc8 and pydocstyle
-	doc8 . docs/
-	pydocstyle copulas/
+lint:
+	ruff check .
+	ruff format .  --check
```
amontanez24, line 89: can we call invoke so we only update in one place

gsheni: Fixed:
https://github.com/sdv-dev/Copulas/pull/407/commits/ff5ced11f1da18dc0f70d12373dfad6e048bcfa5

pyproject.toml
gsheni: Fixed:
https://github.com/sdv-dev/Copulas/pull/407/commits/ff5ced11f1da18dc0f70d12373dfad6e048bcfa5

tests/unit/multivariate/test_gaussian.py
gsheni: The only way to disable this is to use to do something like this:
```
# fmt: off
f(
 a, b, c,
 d, e, f,
)
# fmt: on
```
- Personally, I think it's annoying to have to worry about this and would just let ruff handle it

amontanez24 (approved):

tests/unit/multivariate/test_gaussian.py
amontanez24: Yeah it's not a big deal. I just had to voice my distaste haha

PR #436, opened 2024-10-31
Modularize GaussianMultivariate fit

### LINKED ISSUE #431: Modularize `fit` Method of GaussianMultivariate to Improve Exception Handling
### Problem Description

To improve code reuse, and exception handling, we would like to refactor the fit method of the `GaussianMultivariate` class. The goal is to split it into smaller, more focused methods that handle specific tasks, including better error management during the fitting process.

### Expected behavior

The current fit method performs several tasks within a single function, including column distribution fitting, exception handling, and correlation matrix computation. To enhance extensibility, we will break this method into modular components that isolate responsibilities and handle exceptions more explicitly.

### Expected Steps
1. **Validate Input Data**: Create a separate method to validate the input `X`, ensuring it is a DataFrame. This will handle cases where the input is not in the correct format.
2. **Fit Individual Columns**: Extract the logic for fitting each column to its distribution into its own method. This function should include the logic to handle distribution-specific exceptions.
3. **Handle Exceptions for Fitting**: Move the exception handling (fallback to `GaussianUnivariate` when fitting fails) into a distinct method to improve error handling extensions.
4. **Compute Correlation Matrix**: Leave as it is since its modular already.

pvk-developer:
Resolves #431 
CU-86b2ku10e

### DIFF
--- a/copulas/multivariate/gaussian.py
+++ b/copulas/multivariate/gaussian.py
@@ -70,6 +70,79 @@ class GaussianMultivariate(Multivariate):
 
         return stats.norm.ppf(np.column_stack(U))
 
+    @check_valid_values
+    def fit(self, X):
+        """Compute the distribution for each variable and then its correlation matrix.
+
+        Arguments:
+            X (pandas.DataFrame):
+                Values of the random variables.
+        """
+        LOGGER.info('Fitting %s', self)
+
+        # Validate the input data
+        X = self._validate_input(X)
+        columns, univariates = self._fit_individual_columns(X)
+
+        self.columns = columns
+        self.univariates = univariates
+
+        LOGGER.debug('Computing correlation.')
+        self.correlation = self._get_correlation(X)
+        self.fitted = True
+        LOGGER.debug('GaussianMultivariate fitted successfully')
+
+    def _validate_input(self, X):
+        """Validate the input data."""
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b2ku10e Copulas - Modularize `fit` Method of GaussianMultivariate to Improve Exception Handling #431](https://app.clickup.com/t/86b2ku10e)

copulas/multivariate/gaussian.py
```diff
@@ -100,42 +80,88 @@ def fit(self, X):
         """
         LOGGER.info('Fitting %s', self)
 
+        # Validate the input data
+        X = self._validate_input(X)
+        columns, univariates = self._fit_individual_columns(X)
```
amontanez24, line 85: for simplicity we can probably just call _fit_columns

amontanez24 (approved):
One small comment but looks good! 🥳

copulas/multivariate/gaussian.py
pvk-developer: Addressed in [6d3dc4c](https://github.com/sdv-dev/Copulas/pull/436/commits/6d3dc4c8f467c25fdfcabb9ec422195e88c3dba2)

frances-h (approved):

PR #445, opened 2025-01-09
Create Prepare Release workflow

### LINKED ISSUE #442: Create Prepare Release workflow
### Problem Description

Copulas doesn't have a workflow for preparing the release like some of our other libraries, slowing down releases.

### Expected behavior

- Create a `prepare_release.yml` workflow which (1) creates release notes and (2) runs the static code analysis workflow.
- Delete the `static_code_analysis.yml` workflow

The final workflow should be similar to the one in [SDV-Enterprise](https://github.com/datacebo/SDV-Enterprise/pull/911).

R-Palazzo:
CU-86b2xrmbh
Resolve #442


sdv-team:
Task linked: [CU-86b2xrmbh Copulas - Create Prepare Release workflow #442](https://app.clickup.com/t/86b2xrmbh)

scripts/release_notes_generator.py
```diff
@@ -0,0 +1,152 @@
...
+    return release_notes
+
+
+def update_release_notes(release_notes):
+    """Add the release notes for the new release to the ``HISTORY.md``."""
+    file_path = 'HISTORY.md'
+    with open(file_path, 'r') as history_file:
+        history = history_file.read()
+
+    token = '# History\n\n'
```
fealho, line 134: Enterprise calls this HISTORY, not sure which one we want to use.

fealho (approved):

scripts/release_notes_generator.py
amontanez24: It has to match whatever is actually in the history file, which in this case is '# History'

PR #447, opened 2025-01-09
Support Python 3.13

rwedge:
Addresses #434


rwedge:
We'll need to update the merge rules to expect "unit (macos-latest, 3.13)" instead of "unit (macos-latest, 3.12)"

fealho (approved):
[tox.ini](https://github.com/sdv-dev/Copulas/blob/827851fbcbf4f024531e7883e18dc8220d32c812/tox.ini#L2) also needs to be updated.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py39-lint, py3{8,9,10,11,12}-{readme,unit,end_to_end,numerical,minimum,tutorials}
+envlist = py39-lint, py3{8,9,10,11,12,13}-{readme,unit,end_to_end,numerical,minimum,tutorials}
 
 [testenv]
 skipsdist = false
```
pvk-developer, line 5: We can delete `tox.ini` i think we are not using tox anymore.

rwedge: Maybe that could be a separate issue since we mention tox in the makefile, contributing documentation, etc.

amontanez24 (approved):

PR #473, opened 2025-05-23
Make `BetaUnivariate` reject fitted parameters that obviously don’t match the data 

### LINKED ISSUE #472: `BetaUnivariate` should reject fitted parameters that obviously don’t match the data
### Background
As described in #455, it is possible for the `beta` to fail the fitting process. In this case, the expected behavior is:
1. The `BetaUnivariate` will throw a `FitError`
2. When using this univariate, we implement a try/catch logic and [switch to a fallback distribution](https://github.com/sdv-dev/Copulas/blob/a5f1c5ade7f2d909cc3d8aec36123a6fa6d58784/copulas/multivariate/gaussian.py#L129) instead

This works when `scipy` throws an error. However, there are cases where `scipy` _does not throw an error_ and returns a bad fit. We should catch such cases and throw an error for them.

### Replication
The data below came from [SDV #2542](https://github.com/sdv-dev/SDV/issues/2542). I've pulled out the particular column into this new CSV below.

[example_data.csv](https://github.com/user-attachments/files/20350073/example_data.csv)

The column contains data roughly in the range [1, 7]. When fitting this data, there is no error, but the fitted  parameters clearly don’t make sense:

```python
import pandas as pd
from copulas.univariate import BetaUnivariate

data = pd.read_csv('example_data.csv')
beta = BetaUnivariate()
beta.fit(data)
print(beta._params)
```
```
{
 'loc': 2.1833080938648672,
 'scale': 9.555402199071863e-29,
 'a': 26439403.24259588,
 'b': 28.274298502064756
}
```

Most notably: The `scale` parameter is _so small, it is essentially 0_. This means that distribution is essentially only capable of creating a constant value of ~2.1833. There have been many such examples of this for the Beta distribution specifically.

### Expected Behavior
The `BetaUnivariate` distribution should to check itself after fitting to make sure the parameters make sense. For now, these are two obvious checks we should add:

- The `scale` value should not be close to zero, as in the example above. (Perhaps we can check that it is greater than [EPSILON](https://github.com/sdv-dev/Copulas/blob/main/copulas/utils.py#L11)?)
- The range `[loc, loc+scale]` should have at least _some, non-zero_ overlap with the actual `[data_min, data_max]` range. That is to say:
    - `loc + scale` should be `> data_min` and
    - `loc` should be `< data_max`

If any one of these checks fail, BetaUnivariate should throw a descriptive error explaining that the fitting failed. 
```
Error: Converged parameters for beta distribution have a near-zero range.
```
```
Error: Converged parameters for beta distribution are outside the min/max range of the data.
```

The rest of the code will then adapt to using a fallback.

### Future Work
(Not in scope for now) In the future, we can keep adding/refining the checks for the failure modes that we find. 

In the future, we can also start to investigate how to “fix” the Beta’s fit rather than error out immediately. (For example: Does supplying `floc` and `fscale` help? We’d need to investigate).

fealho:
CU-86b52ccx2, Resolve #472.

### DIFF
--- a/copulas/univariate/beta.py
+++ b/copulas/univariate/beta.py
@@ -4,6 +4,7 @@ import numpy as np
 from scipy.stats import beta
 
 from copulas.univariate.base import BoundedType, ParametricType, ScipyModel
+from copulas.utils import EPSILON
 
 
 class BetaUnivariate(ScipyModel):
@@ -25,9 +26,19 @@ class BetaUnivariate(ScipyModel):
         }
 
     def _fit(self, X):
-        loc = np.min(X)
-        scale = np.max(X) - loc
-        a, b, loc, scale = beta.fit(X, loc=loc, scale=scale)
+        min_x = np.min(X)
+        max_x = np.max(X)
+        a, b, loc, scale = beta.fit(X, loc=min_x, scale=max_x - min_x)
+
+        if loc > max_x or scale + loc < min_x:
+            raise ValueError(
+                'Converged parameters for beta distribution are '
+                'outside the min/max range of the data.'
+            )
+
+        if scale < EPSILON:
+            raise ValueError('Converged parameters for beta distribution have a near-zero range.')
+
         self._params = {'loc': loc, 'scale': scale, 'a': a, 'b': b}
 
     def _is_constant(self):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b52ccx2 Copulas - `BetaUnivariate` should reject fitted parameters that obviously don’t match the data #472](https://app.clickup.com/t/86b52ccx2)

pvk-developer (approved):

PR #498, opened 2025-12-22
Add support for NumPy 2.4.0

### LINKED ISSUE #497: Add support for NumPy 2.4.0
### Environment Details

Please indicate the following details about the environment in which you found the bug:

* Copulas version:
* Python version:
* Operating System:

### Error Description
NumPy 2.4.0 [removed some deprecated functionality](https://numpy.org/doc/stable/release/2.4.0-notes.html#raise-typeerror-on-attempt-to-convert-array-with-ndim-0-to-scalar) that we still rely on, resulting in usage and test errors.

See failures during the Automated Dependency Check [here](https://github.com/sdv-dev/Copulas/pull/496/checks)

frances-h:
Resolve #497

### DIFF
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -31,6 +31,5 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           python -m pip install invoke .[test]
-          python -m pip install "scikit-learn<1.8.0"
     - name: Test with minimum versions
       run: invoke minimum
--- a/tasks.py
+++ b/tasks.py
@@ -21,6 +21,9 @@ COMPARISONS = {
     '<=': operator.le,
     '==': operator.eq,
 }
+EXTERNAL_DEPENDENCY_CAPS = {
+    'scikit-learn': '1.8.0'
+}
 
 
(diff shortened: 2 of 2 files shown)


.github/workflows/minimum.yml
```diff
@@ -31,5 +31,6 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           python -m pip install invoke .[test]
+          python -m pip install "scikit-learn<1.8.0"
```
amontanez24: why is this not a requirement?

amontanez24 (commented):
we do this in SDV to handle the indirect dependencies
https://github.com/sdv-dev/SDV/blob/ad8e4e09bdd01338ea94a60cddd1f06c815273aa/tasks.py#L16

.github/workflows/minimum.yml
frances-h: Scikit-learn is an optional dependency for the tutorials. 1.8.0 recently released and no longer supports older versions of scipy and numpy which is why the minimum tests were originally failing.

fealho (approved):

